PR #3, opened 2021-01-07
Cleanup checkstyle violations

siy:


clstrfsck (approved):
Previously we had specifically excluded org.radix.* packages from checkstyle, as we had intended to replace the over time, but kudos for putting in the work to fix all this.

LukasGasior1 (approved):
LGTM

PR #6, opened 2021-01-11
Feature/fix junit deprecations

clstrfsck:
Replaced JUnit `assertThat(...)` with AssertJ `assertThat(...)`.  Note that the JUnit method is now deprecated, and this change makes our unit tests more uniform.

Added missing Override annotations for various methods in the new interoperability test suite.

Removed some unused loggers.

Slightly modified one use of a method reference into a lambda that will compile with the Eclipse compiler.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# Contributing to radixdlt-java
+# Contributing to radixdlt
 
 ### Table of contents
 - [Code of conduct](#code-of-conduct)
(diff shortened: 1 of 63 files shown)


siy (approved):
LGTM
Also, it would be great to check if Hamcrest is still in use and if not, cleanup dependencies.

Sajjon (approved):
LGTM

LukasGasior1 (approved):

siy (approved):
LGTM

PR #15, opened 2021-01-18
Feature/rpnv1 948 remove version serialization

clstrfsck:

PR #17, opened 2021-01-18
Feature/rpnv1 722 remove destinations from particles

clstrfsck:


Sajjon (approved):
LGTM, but how is the new method:

```java
	@Override
	public Set<EUID> getDestinations() {
		return ImmutableSet.of(this.rri.getAddress().euid());
	}
```

used? Didn't really see the use?

siy (approved):
LGTM, although many files look like repetition of other PR (#15)

clstrfsck:
> LGTM, but how is the new method (`getDestinations()`) used? Didn't really see the use?

It's the same method that was in `Particle` previously.  Previously all subclasses passed the destination EUIDs into `Particle` constructor, which the class stored in a field, and the method `getDestinations()` in `Particle` just returned that field.

This change computes the EUIDs in the subclasses so that there is no explicit field in Particle for this.

clstrfsck:
Will be merged in #20

PR #23, opened 2021-01-19
Cleaned Optional usage

cbisaillon:


radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/ConstraintMachine.java
```diff
@@ -321,7 +321,7 @@ public String toString() {
 						);
 					}
 
-					if (prevUsedData != null && !prevUsedData.isPresent()) {
+					if (prevUsedData.isEmpty()) {
```
talekhinezh, line 324: This class and optional usage definitely needs to be cleaned up but the above statements aren't equivalent as the null state is used to represent first execution of for loop which `prevUsedData.isEmpty()` misses.

talekhinezh (changes requested):

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/ConstraintMachine.java
clstrfsck: Yeah, unfortunately `prevUsedData` is used as a tri-state here, with `null`, empty and present having distinct meanings.

cbisaillon: I understand. What about instead of checking if `prevUsedData` is null, we check if `testInput` is false which doesn't happen during the first iteration ?

LukasGasior1 (approved):

PR #34, opened 2021-01-22
Added error-checked parameters for the script command line:

RayAtRadix:
-n <digits> – the node file number that will be generated.
-p <digits> – the external port number that the service will use.
-f – if this flag is set then the file willl be overwritten without checking if it exists first.

Done to make the documentation a bit easier.

### DIFF
--- a/radixdlt-core/docker/scripts/generate-yml.sh
+++ b/radixdlt-core/docker/scripts/generate-yml.sh
@@ -6,16 +6,40 @@ set -e
 # Where we are run from
 scriptdir=$(dirname "$0")
 
-# Number of validators
-validators=${1:-1}
+while getopts fn:p: flag
+do
+  case "${flag}" in
+    f) force_delete=true;;
+    p) port_param=${OPTARG};;
+    n) validators=${OPTARG};;
+    *) echo "Invalid parameters … exiting"
+       exit 1;;
+  esac
+done
+
+# Being a bit strict. You need  port parameter and a node parameter to continue
+# and they both have to be numbers.
+re='^[0-9]+$'
+
+if ! [[ $port_param =~ $re ]] || ! [[ $validators =~ $re ]]; then
+  echo "Invalid parameter setting"
+  exit 1
+fi
 
 network_name="network_${validators}_nodes"
 file_name="${scriptdir}/../node-${validators}.yml"
 
-if [ -f "${file_name}" ]; then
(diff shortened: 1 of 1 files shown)


radixdlt-core/docker/scripts/generate-yml.sh
```diff
@@ -6,17 +6,41 @@ set -e
...
 
-# Number of validators
-validators=${1:-1}
+while getopts fn:p: flag
+do
+  case "${flag}" in
+    f) force_delete=true;;
+    p) port_param=${OPTARG};;
+    n) validators=${OPTARG};;
+    *) echo "Invalid parameters … exiting"
```
cbisaillon, line 15: I think it would be a good idea to add a -h parameter to explain the syntax.
Then instead of printing "Invalid parameters", print the same message as -h

RayAtRadix: Makes a lot of sense. Yup, I'll do that.

radixdlt-core/docker/scripts/generate-yml.sh
```diff
@@ -53,7 +104,7 @@ do
     echo "    networks:">>${file_name}
     echo "      - $network_name">>${file_name}
     echo "    ports:">>${file_name}
-    echo "      - \"80${p0}:8080\"">>${file_name}
+    echo "      - \"${port_param}${p0}:8080\"">>${file_name}
```
clstrfsck, line 107: Suggest that this would be more predictable if `port_param` was the initial port number, and that subsequent nodes used subsequent port numbers, rather than dividing things into two parts here.

For example, it should be the case that using `-p 8080 -n 2` would result in ports 8080 and 8081 being generated.  In order to do this at the moment, would need to use `-p 80 -n 2` which is kind of confusing.  Also, it is not currently possible to generate (for example) configs that use ports 8000, 8001 etc, given the way that `p0` is set up here now.

clstrfsck (changes requested):
LGTM

radixdlt-core/docker/scripts/generate-yml.sh
RayAtRadix: Yep, good point.

RayAtRadix:
If there are no objections, I'll hit the merge pull request button :-)

clstrfsck (approved):
Looks good.

PR #37, opened 2021-01-23
Fixed null error if validatorSet was not set with BFTBuilder

cbisaillon:
Would it be better to just throw an exception if validatorSet is null ?


cbisaillon:
Why use Builder instead of a normal constructor if all the arguments are required to be non null?

talekhinezh:
> Why use Builder instead of a normal constructor if all the arguments are required to be non null?

Yes, this is a good point. Main reason is that this code is from an older architecture and hasn't been properly refactored yet.

cbisaillon:
Is it ok if I refactor to a constructor ?

cbisaillon:
I'll close this and refactor the class to use a constructor instead of Builder

PR #41, opened 2021-01-28
Bugfix/rpnv1 962 exponential message growth2

clstrfsck:
There are four major things happening here:

- (BFTSync) Ensure that highest QC request is prioritised
- (BFTSync) Rate limit outbound requests and importantly request timeouts
- (LocalSyncServiceAccumulatorProcessor) continue requesting further updates without waiting for a timeout if sync is incomplete
- (EpochManager) ensure that out-of-epoch error responses result in a epoch sync for the current epoch


clstrfsck:
There are four major things happening here:
- (BFTSync) Reduce sync request growth by pruning response errors that result in a new request
- (BFTSync) Rate limit outbound requests and importantly request timeouts
- (LocalSyncServiceAccumulatorProcessor) continue requesting further updates without waiting for a timeout if sync is incomplete
- (EpochManager) ensure that out-of-epoch error responses result in a epoch sync for the current epoch

siy (approved):
LGTM

LukasGasior1 (approved):

PR #47, opened 2021-02-01
Experiment/new persistence layer implementation

siy:


siy:
Experiment results shows that LSM-based K-V stores are performing worse in our use case.

PR #58, opened 2021-02-05
Make universe validation read resource file directly

clstrfsck:
instead of (possibly old) default.config in default directory


talekhinezh (approved):

siy (approved):
LGTM

PR #64, opened 2021-02-16
Feature/refactor events

talekhinezh:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/environment/Dispatchers.java
```diff
@@ -79,14 +80,26 @@ private Dispatchers() {
...
 
 		@Override
 		public ScheduledEventDispatcher<T> get() {
-			return environmentProvider.get().getScheduledDispatcher(c);
+			Environment e = environmentProvider.get();
+			if (c != null) {
+				return e.getScheduledDispatcher(c);
+			} else {
+				return e.getScheduledDispatcher(t);
+			}
```
siy, line 102: I think `c` and `t` should have more informative names, as catching small (one letter) difference between return statements is rather inconvenient. Also, this should help to realize that actually here can be invoked two different methods rather than one with different parameters.

siy (approved):
LGTM with one comment

PR #65, opened 2021-02-16
Add backpressure buffer to TCP/UDP channels processor

LukasGasior1:
Without explicit buffer the stream can explode with an exception:

```
Exception in thread "PeerManager" io.reactivex.rxjava3.exceptions.OnErrorNotImplementedException: The exception was not handled due to missing onError handler in the subscribe() method call.
Further reading: https://github.com/ReactiveX/RxJava/wiki/Error-Handling | io.reactivex.rxjava3.exceptions.MissingBackpressureException: Could not emit value due to lack of requests
```

happens during app startup in kubernetes network.


siy (approved):
LGTM

talekhinezh (approved):

PR #68, opened 2021-02-19
Allow to replace a timeout vote

LukasGasior1:


siy (approved):
LGTM once non-mocked test will be added

PR #72, opened 2021-02-24
RPNV1-1036  Add DB location to default.config.envsubst

rvelaz:
Allow specifying path for ledger DB


radixdlt-core/docker/config/default.config.envsubst
```diff
@@ -13,3 +13,4 @@ chaos.enable=$RADIXDLT_CHAOS_ENABLE
 
 host.ip=$RADIXDLT_HOST_IP_ADDRESS
 cp.port=$RADIXDLT_CP_PORT
+db.location=$RADIXDLT_DB_LOCATION
```
shambupujar, line 16: You may need to add a default environment variable in dockerfile (radixdlt-core/docker/Dockerfile.core) . Otherwise it will fail when someone doesn't setup the environment on docker run

shambupujar (changes requested):

radixdlt-core/docker/config/default.config.envsubst
rvelaz: done

shambupujar (approved):

LukasGasior1 (approved):
LGTM

PR #80, opened 2021-03-04
Chaos testing on testnets

Theoklitos:
Added a chaos experiment which should run against a testnet (milestonenet).

The test is found in the [ChaosTest ](https://github.com/radixdlt/radixdlt/blob/task/add_chaos_to_testnet/radixdlt-regression/system-tests/src/test/java/com/radixdlt/test/chaos/ChaosTests.java)class. It works like this:

- Make sure we got liveness
- Perform several actions against the network
- Make sure we still got liveness

The actions do things like add network latency or restart containers. They can be found under [com.radixdlt.test.chaos.actions](https://github.com/radixdlt/radixdlt/tree/task/add_chaos_to_testnet/radixdlt-regression/system-tests/src/test/java/com/radixdlt/test/chaos/actions). They use ansible and [JSch](http://www.jcraft.com/jsch/) to do stuff to the network.

They are meant to be run from Jenkins periodically, which is why each action has a random chance of getting applied.

### DIFF
--- a/radixdlt-regression/system-tests/build.gradle
+++ b/radixdlt-regression/system-tests/build.gradle
@@ -33,7 +33,6 @@ dependencies {
     implementation 'io.reactivex.rxjava2:rxjava'
     implementation 'org.json:json'
 
-    //testImplementation 'com.jcraft:jsch:0.1.53'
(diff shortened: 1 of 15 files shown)


siy (approved):
LGTM

PR #81, opened 2021-03-04
Working ECIES with DH addition

Sajjon:
Working POC (with test) of ["Cyon" ECIES](https://crypto.stackexchange.com/q/88083/60476) using [AES GCM](https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Galois/counter_(GCM)) instead of [AES CBC](https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Cipher_block_chaining_(CBC)) + [HMAC](https://en.wikipedia.org/wiki/HMAC).

Note, we replace two steps AES-CBC and HMAC with just AES-GCM, since GCM has authentication built in.

This is not meant to be a final implementation, but could easily refactored and tidied up to be! It proves that we have two cross library compatible implementations of this new "virtually symmetric" encryption scheme. By "virtually symmetric" I mean that it is still Elliptic Curve based asymmetric encryption, using symmetric encryption (AES) internally, but thanks to DiffieHellman key exchange + extra safety thanks to the addition of an ephemeral point, **_we enable both recipient AND sender to be able to decrypt the message_** - without any application layer constructs.

[Corresponding POC in Swift (Cyons lib "EllipticCurveKit")](https://github.com/Sajjon/EllipticCurveKit/pull/18), note exact same test vector as implemented in this PR on [those lines](https://github.com/Sajjon/EllipticCurveKit/blob/7fcb193d976c39ce3f3202be79cb10406f081e8e/Tests/EllipticCurveKitTests/Crypto/ECIESTests.swift#L126-L133).

### DIFF
--- a/radixdlt-java-common/src/main/java/com/radixdlt/crypto/encryption/ECIES.java
+++ b/radixdlt-java-common/src/main/java/com/radixdlt/crypto/encryption/ECIES.java
@@ -26,9 +26,13 @@ import com.radixdlt.crypto.ECPublicKey;
 import com.radixdlt.crypto.HashUtils;
 import com.radixdlt.crypto.exception.MacMismatchException;
 import com.radixdlt.crypto.exception.ECIESException;
+import org.bouncycastle.jcajce.provider.digest.SHA256;
 import org.bouncycastle.math.ec.ECPoint;
 
+import javax.crypto.Cipher;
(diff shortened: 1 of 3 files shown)


Sajjon:
Encryption is not needed in Java lib, closing this PR.

PR #104, opened 2021-03-12
Cleanup message preprocessor

LukasGasior1:


radixdlt-core/radixdlt/build.gradle
```diff
@@ -149,6 +149,13 @@ dependencies {
     implementation 'com.stijndewitt.undertow.cors:undertow-cors-filter'
     implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml'
 
+    implementation 'io.vavr:vavr:1.0.0-alpha-3'
```
LukasGasior1, line 152: this is for `Either`

radixdlt-core/radixdlt/build.gradle
```diff
@@ -149,6 +149,13 @@ dependencies {
...
+
+    compileOnly 'org.projectlombok:lombok:1.18.16'
```
LukasGasior1, line 154: and this is for friendly value classes, like this one:
```
	@Value(staticConstructor = "of")
	static class MessageProcessingError {
		String msg;
	}
```

radixdlt-core/radixdlt/build.gradle
siy: Not sure that we actually need `Either`. In most cases we need something like 'Result' rather than `Either`. And such a class is rather simple, so we might prefer to implement them by ourselves rather than add more dependencies.

LukasGasior1: union types are very common, from my experience it's quite useful to have one around

radixdlt-java-common/src/main/java/com/radixdlt/utils/functional/FailureType.java
```diff
@@ -0,0 +1,24 @@
...
+ * either express or implied.  See the License for the specific
+ * language governing permissions and limitations under the License.
+ */
+
+package com.radixdlt.utils.functional;
+
+/**
+ * Interface for various failure types.
+ */
+public interface FailureType {
```
talekhinezh, line 23: We shouldn't introduce generic marker interfaces unless there's a clear need to.

talekhinezh (changes requested):

talekhinezh (approved):

PR #109, opened 2021-03-15
Feature/consolidate atom

talekhinezh:


radixdlt-engine/src/main/java/com/radixdlt/atom/Atom.java
```diff
@@ -77,13 +79,21 @@
...
+		this(List.of(pg), Map.of(), null);
+	}
+
+	public Atom(List<ParticleGroup> pgs) {
+		this(pgs, Map.of(), null);
+	}
+
 	public Atom(List<ParticleGroup> particleGroups, Map<EUID, ECDSASignature> signatures, @Nullable String message) {
 		Objects.requireNonNull(particleGroups, "particleGroups is required");
 		Objects.requireNonNull(signatures, "signatures is required");
```
siy, line 99: It's a good practice to avoid exceptions in the constructor. For this purpose most often used approach is the `single constructor + several factory methods` instead of several constructors with different set of parameters. I think we should continue use of factory methods as it was before.

siy (approved):
LGTM with one comment

LukasGasior1 (approved):

radixdlt-engine/src/main/java/com/radixdlt/atom/Atom.java
talekhinezh: yes agreed, this will be fixed in upcoming PR regarding atom refactor

PR #124, opened 2021-03-19
Client API DB

siy:


radixdlt-core/radixdlt/src/test/java/com/radixdlt/client/store/berkeley/BerkeleyClientApiStoreTest.java
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	@Test
+	public void tokenBalancesAreReturned() {
+		var particles = List.of(SpunParticle.up(stake(UInt256.TWO)), SpunParticle.up(transfer(UInt256.NINE)));
+		var clientApiStore = prepareApiStore(particles);
+
+		clientApiStore.getTokenBalances(OWNER)
+			.onSuccess(list -> {
+				assertEquals(2, list.size());
```
Sajjon, line 76: I **think** this test is double incorrect, but I might have misunderstood something?:

# 1. Balances should be reduced/aggregated/merged *per token* (RRI)
Why does this test return a list of length 2? Both your two helper methods `stake(UInt256 amount)` and `transfer(UInt256 amount)` respectively use the **same** `RRI`, so the list should really just have contained one entry

We should NOT return these two balance entries

```
[
     { amount: 3, rri: "ALICE_TOKEN" }, 
     { amount: 5, rri: "ALICE_TOKEN" }, 
]
``` 

for these particles (pseudocode):

```
[
	SpunParticle {
		spin: UP,
		particle: {
			type: TransferrableTokensParticle,
			rri: ALICE_TOKEN
			amount: 3
		}
	},
	SpunParticle {
		spin: UP,
		particle: {
			type: TransferrableTokensParticle,
			rri: ALICE_TOKEN
			amount: 5
		}
	},
]
```

Instead, we should return the single item (pseudocode):

```
[
     { amount: 8, rri: "ALICE_TOKEN" }, 
]
```

Being the calculated total balance of the `"ALICE_TOKEN"`.

# 2. Staked amount should be _subtracted_ 
All staked tokens should considered as "spent", or rather, considered as _not  **immediately** spendable_

So this test should be update to this:

```java
clientApiStore.getTokenBalances(OWNER)
			.onSuccess(list -> {
				assertEquals(1, list.size());
				assertEquals(UInt256.NINE, list.get(0).getAmount());
				assertEquals(TOKEN, list.get(0).getRri());
			})
```

Just a **single entry per token (RRI)** and having the amount `NINE`, since that is what was transferred to the owner,, the `TWO` is staked, so not spendable.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/store/berkeley/BerkeleyClientApiStore.java
```diff
@@ -0,0 +1,296 @@
...
+	public BerkeleyClientApiStore(DatabaseEnvironment dbEnv, LedgerEntryStore store, Serialization serialization) {
+		this.dbEnv = dbEnv;
+		this.store = store;
+		this.serialization = serialization;
+
+		open();
+	}
+
+	@Override
+	public Result<List<TokenBalance>> getTokenBalances(RadixAddress address) {
```
Sajjon, line 82: This should only return one entry per token (RRI), with staked amount subtracted, only immediately spendable tokens :).

Sajjon (changes requested):
Your code is beautiful @siy ! A joy to read, really! Great job! :D 

However, I'm pretty sure `getTokenBalances` is incorrectly implemented. Have a look at my inline comments :).

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/store/berkeley/BerkeleyClientApiStore.java
```diff
@@ -0,0 +1,296 @@
...
+public class BerkeleyClientApiStore implements ClientApiStore {
+	private static final Logger log = LogManager.getLogger();
+
+	private static final String EXECUTED_TRANSACTIONS_DB = "radix.executed_transactions_db";
+	private static final String BALANCE_DB = "radix.balance_db";
+
+	private final DatabaseEnvironment dbEnv;
+	private final LedgerEntryStore store;
+	private final Serialization serialization;
+	private final StackingCollector<SpunParticle> particleCollector = StackingCollector.create();
```
talekhinezh, line 66: Better if this were outside of ClientApiStore so that event/thread handling can be independent from client api db. (For example would want to integrate this with Deterministic tests at some point all of which run on a single thread.)

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/store/berkeley/StackingCollector.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+
+/**
+ * The purpose of this data structure is to collect elements (using {@link #push(Object)} method) and then
+ * process collected elements at once using {@link #consumeCollected(Consumer)} method. While processing takes place
+ * new elements can be stored (they will be processed during next call to {@link #consumeCollected(Consumer)}).
+ */
+public final class StackingCollector<T> {
```
talekhinezh, line 28: Perhaps better to use RxJava3 primitives to handle async boundary logic. Would then be able to integrate with our current dispatcher framework.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/LedgerEntryStoreView.java
```diff
@@ -45,6 +45,15 @@
...
 	Optional<Atom> get(AID aid);
 
+	/**
+	 * Gets the atom directly from commit log.
+	 *
+	 * @param address
+	 * 			Address to read atom from.
+	 * @return the atom stored at given address
+	 */
+	Optional<Atom> get(long address);
```
talekhinezh, line 55: This isn't used. Please delete.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/berkeley/BerkeleyLedgerEntryStore.java
```diff
@@ -225,6 +250,34 @@ public LedgerEntryStoreResult store(
...
+			var status = cursor.getFirst(key, value, DEFAULT);
+
+			while (status == SUCCESS) {
+				if (value.getData().length > 0) {
+					var particleBytes = Arrays.copyOfRange(value.getData(), EUID.BYTES, value.getData().length);
+					var particle = deserializeOrElseFail(particleBytes, Particle.class);
+					particleConsumer.accept(SpunParticle.up(particle));
+				} else {
+					//TODO: eventually we may want to store down particles as well
+					particleConsumer.accept(SpunParticle.down(VOID_PARTICLE));
```
talekhinezh, line 267: No won't be storing down particles in this part of the DB. Balance should only be dependent on UP particles so this is not needed.

talekhinezh (changes requested):

radixdlt-core/radixdlt/src/test/java/com/radixdlt/client/store/berkeley/BerkeleyClientApiStoreTest.java
```diff
@@ -0,0 +1,102 @@
...
+	public TemporaryFolder folder = new TemporaryFolder();
+
+	@Before
+	public void setUp() {
+		environment = new DatabaseEnvironment(folder.getRoot().getAbsolutePath(), 0);
+	}
+
+	@Test
+	public void tokenBalancesAreReturned() {
+		var particles = List.of(SpunParticle.up(stake(UInt256.TWO)), SpunParticle.up(transfer(UInt256.NINE)));
```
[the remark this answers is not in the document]
siy: Fixed

radixdlt-core/radixdlt/src/test/java/com/radixdlt/client/store/berkeley/BerkeleyClientApiStoreTest.java
siy: Fixed

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/store/berkeley/BerkeleyClientApiStore.java
siy: Fixed

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/store/ClientApiStore.java
```diff
@@ -0,0 +1,31 @@
...
+import com.radixdlt.utils.functional.Result;
+
+import java.util.List;
+
+/**
+ * High level JSON RPC client API store.
+ */
+public interface ClientApiStore {
+
+	Result<List<TokenBalance>> getTokenBalances(RadixAddress address);
```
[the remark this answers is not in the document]
siy: So far there is no such requirement in the document. Also, since balance is calculated per token, for average user the list should not be too big. Nevertheless, we may decide to add pagination in the future.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/store/berkeley/StackingCollector.java
siy: This class is just a thread-safe temporary storage.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/LedgerEntryStoreView.java
siy: Yep, it's a leftover from previous design. Fixed.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/berkeley/BerkeleyLedgerEntryStore.java
siy: To calculate balances properly I need to have both, UP and DOWN particles. Also, beside balances there is a need to calculate token supply. The only method I see here is to handle unallocated tokens particle. Again, both, UP and DOWN particles are necessary to calculate supply.

Actually main issue here is the retrieving historical data during DB rebuild. Under normal circumstances full body of UP and DOWN particles are available, so there is no problem to calculate all things mentioned above.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/store/berkeley/BerkeleyClientApiStore.java
```diff
@@ -0,0 +1,296 @@
...
+			if (System.getProperty("db.check_integrity", "1").equals("1")) {
+				//TODO: Implement recovery, basically should be the same as fresh DB handling
+			}
+
+			if (executedTransactionsDatabase.count() == 0) {
+				//Fresh DB, rebuild from log
+				rebuildDatabase();
+			}
+
+			store.onParticleCommit(this::newParticle);
```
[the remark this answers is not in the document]
siy: I see in your PR related change which enables subscription on atom commit events. Once your PR will be merged, I'll refactor this part to use that subscription.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/store/berkeley/BerkeleyClientApiStore.java
siy: I've removed scheduler and added method to explicitly flush this temporary storage into persistent DB. This basically removes all internal asynchronous processing and makes whole processing deterministic. This way we can integrate client API DB with any external scheduling mechanism.

Sajjon (approved):
LGTM

PR #131, opened 2021-03-23
Feature/fix atom id

talekhinezh:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineMempool.java
```diff
@@ -101,62 +100,55 @@ public void add(Command command) throws MempoolRejectedException {
...
-				if (command != null) {
-					Atom atom;
-					try {
-						atom = DefaultSerialization.getInstance().fromDson(command.getPayload(), Atom.class);
-					} catch (DeserializeException e) {
-						throw new IllegalStateException();
-					}
-
-					removed.add(Pair.of(atom, new RadixEngineMempoolException(
+				if (toRemove != null && !toRemove.equals(atom)) {
```
siy, line 121: Perhaps it worth to reverse the comparison: `atom.equals(toRemove)`. This should make comparison safe even if `data.remove(command)` will return `null`.

siy (approved):
LGTM with few comments

PR #132, opened 2021-03-23
Add deprecation warnings on java lib

Theoklitos:
Added `@Deprecation` annotation and `@deprecation` javadoc on most public methods in RadixApplicationAPI.


siy (approved):
LGTM

PR #146, opened 2021-03-29
Feature/local substate shutdown

talekhinezh:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/berkeley/BerkeleyLedgerEntryStore.java
```diff
@@ -519,11 +521,11 @@ private void downParticle(com.sleepycat.je.Transaction txn, byte[] substateId) {
 		final var downedParticle = entry();
 		var status = particleDatabase.get(txn, entry(substateId), downedParticle, DEFAULT);
 		if (status != SUCCESS) {
-			throw new IllegalStateException("Downing particle does not exist " + substateId);
+			throw new IllegalStateException("Downing particle does not exist " + Hex.toHexString(substateId));
```
siy, line 524: It might be more convenient to add `toHexString` method to `SubstateId` class.

radixdlt-engine/src/main/java/com/radixdlt/atom/Atom.java
```diff
@@ -58,7 +58,7 @@
 	@DsonOutput({Output.ALL})
 	private final ECDSASignature signature;
 
-	private final ImmutableList<REInstruction> instructions;
+	private final List<REInstruction> instructions;
```
siy, line 61: I really like this change. I think we should get rid of ImmutableList/Map/Set/etc. in field and parameter declarations.

siy (approved):
LGTM

PR #172, opened 2021-04-12
Refactor Constraint Machine

talekhinezh:
* Refactor token mint/burn by removing UnallocatedTokensParticle
* Refactor handling of high level api side parsing, utilizing Constraint Machine logic rather than manually parsing
* Changed Balance entries to use UInt384 rather than UInt256 due to the change in mints/burns


siy (approved):
LGTM

LukasGasior1 (approved):
LGTM

PR #178, opened 2021-04-15
Use uncompressed representation for ECPublicKey equality

LukasGasior1:
# Pull Request Template

## Title
Use uncompressed representation for ECPublicKey equality

## Description

Previously, if we initiated two instances of a public key of the same EC point with compressed and uncompressed byte representation (using `ECPublicKey.fromBytes`) we would end up with two unequal objects. This PR changes it so that they're equal. Precisely: two public keys are equal if their uncompressed byte representation is equal. It also adds a validation that the bytes passed actually represent a valid point. This change also revealed that some classes were implicitly relying and being given a key with compressed bytes representation (RadixAddress for example).

### DIFF
--- a/radixdlt-java-common/src/main/java/com/radixdlt/crypto/ECKeyPair.java
+++ b/radixdlt-java-common/src/main/java/com/radixdlt/crypto/ECKeyPair.java
@@ -70,7 +70,7 @@ public final class ECKeyPair implements Signing<ECDSASignature> {
 			ECPrivateKeyParameters privParams = (ECPrivateKeyParameters) keypair.getPrivate();
 			ECPublicKeyParameters pubParams = (ECPublicKeyParameters) keypair.getPublic();
 
-			final ECPublicKey publicKey = ECPublicKey.fromBytes(pubParams.getQ().getEncoded(true));
+			final ECPublicKey publicKey = new ECPublicKey(pubParams.getQ());
 			byte[] privateKeyBytes = ECKeyUtils.adjustArray(privParams.getD().toByteArray(), BYTES);
 			ECKeyUtils.validatePrivate(privateKeyBytes);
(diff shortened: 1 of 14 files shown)


radixdlt-java-common/src/test/java/com/radixdlt/identifiers/RadixAddressTest.java
```diff
@@ -52,7 +52,7 @@ public void when_an_address_is_created_with_same_string__they_should_be_equal_an
 
 	@Test
 	public void address_from_key_and_magical() throws PublicKeyException {
-		String publicKeyHexString = "03000102030405060708090A0B0C0D0E0F101112131415161718191A1B1C1D1E1F";
```
LukasGasior1, line 55: changed because these bytes didn't represent a valid ec point (which previously wasn't validated)

radixdlt-java-common/src/main/java/com/radixdlt/crypto/ECKeyPair.java
```diff
@@ -70,7 +70,7 @@ public static ECKeyPair generateNew() {
 			ECPrivateKeyParameters privParams = (ECPrivateKeyParameters) keypair.getPrivate();
 			ECPublicKeyParameters pubParams = (ECPublicKeyParameters) keypair.getPublic();
 
-			final ECPublicKey publicKey = ECPublicKey.fromBytes(pubParams.getQ().getEncoded(true));
+			final ECPublicKey publicKey = new ECPublicKey(pubParams.getQ());
```
siy, line 73: I think it's better to have factory method for this rather than call constructor directly.

siy (approved):
LGTM

michalzajda (approved):

PR #182, opened 2021-04-16
Fix Pre-signed blob

talekhinezh:
* Implement presigned transaction as a single byte blob
* Split Constraint Machine to stateless and stateful verification


siy (approved):
LGTM

PR #197, opened 2021-04-20
Validators endpoint

siy:
Initial implementation of "validators" JSON RPC method. Things which are not yet working:
- Owner and validator are always the same. Separation of these things not yet implemented in consensus code.
- Owner's own stake is not calculated.

### DIFF
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/api/ValidatorInfoDetails.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/api/ValidatorInfoDetails.java
@@ -26,7 +26,7 @@ import static org.radix.api.jsonrpc.JsonRpcUtil.jsonObject;
 
 import static java.util.Objects.requireNonNull;
 
-public class ValidatorDetails {
+public class ValidatorInfoDetails {
 	private final RadixAddress address;
 	private final RadixAddress owner;
 	private final String name;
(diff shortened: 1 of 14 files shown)


michalzajda (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/api/ValidatorInfoDetails.java
```diff
@@ -0,0 +1,88 @@
...
+
+import com.radixdlt.identifiers.RadixAddress;
+import com.radixdlt.utils.UInt256;
+
+import static org.radix.api.jsonrpc.JsonRpcUtil.jsonObject;
+
+import static java.util.Objects.requireNonNull;
+
+public class ValidatorInfoDetails {
+	private final RadixAddress address;
```
Sajjon, line 30: better change this to `String address` for now? because it will be `ValidatorAddress` class later, so letting it be `RadixAddress` is just misleading and breaks API? Will will be sending arbitrary string from JS

Sajjon (changes requested):
Think maybe we should change `RadixAddress address` to `String address`?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/api/ValidatorInfoDetails.java
talekhinezh: or `ECPublicKey`

PR #209, opened 2021-04-22
Refactor p2p layer

LukasGasior1:
## Title
Refactor p2p layer

## Description

Introduce encrypted, RLPx-like peer to peer communication.

Dependent PR - https://github.com/radixdlt/radixdlt-iac/pull/139  and should be merged at the same time


radixdlt-core/radixdlt/src/main/java/com/radixdlt/keys/PersistedBFTKeyManager.java
```diff
@@ -68,4 +68,8 @@ public ECDSASignature sign(byte[] hash) {
 	public BFTNode self() {
 		return BFTNode.create(ecKeyPair.getPublicKey());
 	}
+
+	public ECKeyPair getKeyPair() {
```
talekhinezh, line 72: I don't think we should be exposing the private key in this way, better would be if exposed a Signer instead.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/keys/PersistedBFTKeyModule.java
```diff
@@ -51,6 +52,12 @@ ECPublicKey key(@Self BFTNode bftNode) {
 		return bftNode.getKey();
 	}
 
+	@Provides
+	@Self
+	ECKeyPair ecKeyPair(PersistedBFTKeyManager keyManager) {
```
talekhinezh, line 57: I don't think we should be exposing the private key in this way, better would be if exposed a Signer instead.

talekhinezh: E.g. the interface we would have if we were to use a private key in hardware

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PModule.java
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	@Provides
+	public P2PConfig p2pConfig() {
+		return P2PConfig.fromRuntimeProperties(this.properties);
+	}
+
+	@Provides
+	@Self
+	public RadixNodeUri selfUri(@Self ECPublicKey selfKey, HostIp hostIp, P2PConfig p2pConfig) {
```
talekhinezh, line 96: Not a fan of having to provide a whole `P2PConfig` to just get the port which is why I like annotations to single values better. But this is fine...for now.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PeerEvent.java
```diff
@@ -0,0 +1,128 @@
...
+ * language governing permissions and limitations under the License.
+ */
+
+package com.radixdlt.network.p2p;
+
+import com.radixdlt.network.p2p.transport.PeerChannel;
+
+import java.util.Objects;
+
+public interface PeerEvent {
```
talekhinezh, line 24: Some javadoc would be helpful, what's the purpose of this marker interface? Why not just treat each PeerEvent as a separate class without a shared marker interface?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/RadixNodeUri.java
```diff
@@ -0,0 +1,105 @@
...
+		return nodeId;
+	}
+
+	@JsonValue
+	private byte[] getSerializedValue() {
+		return getUriString().getBytes(StandardCharsets.UTF_8);
+	}
+
+	private String getUriString() {
+		return String.format("radix://%s@%s:%s", nodeId.getPublicKey().toBase58(), host, port);
```
talekhinezh, line 79: Like mentioned lets use bech32

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBook.java
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Manages known peers network addresses and their metadata.
+ */
+@Singleton
+public final class AddressBook {
+	private static final Comparator<AddressBookEntry.PeerAddressEntry> entryComparator = (a, b) -> {
+		final var aLastSuccess = a.getLastSuccessfulConnection().orElse(Instant.MIN);
+		final var bLastSuccess = b.getLastSuccessfulConnection().orElse(Instant.MIN);
+		return -aLastSuccess.compareTo(bLastSuccess);
```
talekhinezh, line 46: Seems like this comparator basically ignores old entries? Or is there a way to get back to older entries?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PeerEvent.java
LukasGasior1: just so we can group related events into a single event dispatcher/processor

michalzajda (approved):

PR #213, opened 2021-04-23
Account address update

talekhinezh:


radixdlt-core/radixdlt/src/test/java/com/radixdlt/client/handler/HighLevelApiHandlerTest.java
```diff
@@ -74,13 +74,12 @@
...
 public class HighLevelApiHandlerTest {
-	private static final byte MAGIC = (byte) 0;
-	private static final String KNOWN_ADDRESS_STRING = "JH1P8f3znbyrDj8F4RWpix7hRkgxqHjdW2fNnKpR3v6ufXnknor";
-	private static final RadixAddress KNOWN_ADDRESS = RadixAddress.from(KNOWN_ADDRESS_STRING);
-
-	private static final RadixAddress V1 = new RadixAddress(MAGIC, ECKeyPair.generateNew().getPublicKey());
-	private static final RadixAddress V2 = new RadixAddress(MAGIC, ECKeyPair.generateNew().getPublicKey());
-	private static final RadixAddress V3 = new RadixAddress(MAGIC, ECKeyPair.generateNew().getPublicKey());
+	private static final ECPublicKey PUB_KEY = ECKeyPair.generateNew().getPublicKey();
+	private static final REAddr ACCOUNT_ADDR = REAddr.ofPubKeyAccount(PUB_KEY);
```
Sajjon, line 78: slighlt confusing names here, the actual AccountAddress is called `ADDRESS` and the ("abstract") REAddr is called `ACCOUNT_ADDR`.  Not important but I would slightly change that?

radixdlt-core/radixdlt/src/test/java/com/radixdlt/client/AccountAddressTest.java
```diff
@@ -0,0 +1,109 @@
...
+			"deadbeef", "brx1qspsel805pa0nhtdhemshp7hm0wjcvd60a8ulre6zxtd2qh3x4smq3sraak9a",
+			"deadbeefdeadbeef", "brx1qsp7gnv7g60plkk9lgskjghdlevyve6rtrzggk7x3fwmp4yfyjza7gcumgm9f",
+			"bead", "brx1qsppw0z477r695m9f9qjs3nj2vmdkd3rg4mfx7tf5v0gasrhz32jefqwxg7ul",
+			"aaaaaaaaaaaaaaaa", "brx1qspqsfad7e5k2k9agq74g40al0j9cllv7w0ylatvhy7m64wyrwymy5g7md96s"
+		)
+	);
+
+	private final BiMap<String, String> reAddrToAccountAddress = HashBiMap.create(
+		Map.of(
+			"03" + "03".repeat(26), "brx1qvpsxqcrqvpsxqcrqvpsxqcrqvpsxqcrqvpsxqcrqvpsytf8zx",
```
Sajjon, line 47: uhm, this is 26 bytes only, this is the recipe for RRIs, should not work with AccountAddress, right? 

If it does, we allow for AccountAddresses NOT containing PublicKey and this breaks stuff, e.g. encryption wont work. I thought we where going to enforce the compressed public key as data for now, and ONLY allow that, so that we on JS side ALWAYS can retrieve the public key from an AccountAddress.

So hmm... I would be happy to see an update to impl that enforces 33 bytes data (compressed public key) and.

Sajjon (changes requested):
I would be happy to see an update to implementation that enforces 33 bytes data (compressed public key) for AccountAddresses so that we do not break encryption/JS implementation. At least for now, and we might revisit later.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/AccountAddress.java
```diff
@@ -0,0 +1,82 @@
...
+			throw new DeserializeException("Could not decode string: " + v, e);
+		}
+
+		if (!bech32Data.hrp.equals("brx")) {
+			throw new DeserializeException("hrp must be vb but was " + bech32Data.hrp);
+		}
+
+		try {
+			var addrBytes = fromBech32Data(bech32Data.data);
+			return REAddr.of(addrBytes);
```
Sajjon, line 69: can we throw an error if addrBytes.length != 33 ?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/AccountAddress.java
```diff
@@ -0,0 +1,82 @@
...
+	private static byte[] toBech32Data(byte[] bytes) {
+		return Bits.convertBits(bytes, 0, bytes.length, 8, 5, true);
+	}
+
+	private static byte[] fromBech32Data(byte[] bytes) {
+		return Bits.convertBits(bytes, 0, bytes.length, 5, 8, false);
+	}
+
+	public static String of(REAddr addr) {
+		var convert = toBech32Data(addr.getBytes());
```
Sajjon, line 51: can we throw an exception if `addr.getBytes().length != 33`?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/AccountAddress.java
```diff
@@ -0,0 +1,82 @@
...
+import com.radixdlt.utils.functional.Result;
+import org.bitcoinj.core.AddressFormatException;
+import org.bitcoinj.core.Bech32;
+
+/**
+ * Bech-32 encoding/decoding of account addresses.
+ *
+ * The human-readable part is "rdx" for mainnet, "brx" for betanet.
+ *
+ * The data part is a conversion of the 1-34 byte Radix Engine address
```
Sajjon, line 33: the unit test vector in `AccountAddressTest.java`:
```java
"03" + "03".repeat(26), "brx1qvpsxqcrqvpsxqcrqvpsxqcrqvpsxqcrqvpsxqcrqvpsytf8zx",
```

contradicts the statement here : `1-34 byte Radix Engine address`, because that test is only 26 bytes, right? I wold like to see 33 bytes enforced.

radixdlt-core/radixdlt/src/test/java/com/radixdlt/client/AccountAddressTest.java
```diff
@@ -0,0 +1,109 @@
...
+			"04" + "02".repeat(33), "brx1qspqyqszqgpqyqszqgpqyqszqgpqyqszqgpqyqszqgpqyqszqgpqyqs7cr9az"
```
Sajjon, line 48: Ha! I'm baffled that this is indeed a valid public key! Where did you find it? On uncompressed form it is:
`040202020202020202020202020202020202020202020202020202020202020202415456f0fc01d66476251cab4525d9db70bfec652b2d8130608675674cde64b2`

Sajjon (commented):
Forgot to say, good job! :)

LukasGasior1 (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/AccountAddress.java
talekhinezh: Wording is confusing. 1-34 bytes refers to the RadixEngineAddress which is a low level address, compared to the bech32 account address which maps to a RadixEngineAddress.

PR #239, opened 2021-04-26
Use strict check that the timeout event matches pending sync request

LukasGasior1:
# Pull Request Template

## Title
Use strict check that the timeout event matches pending sync request

## Description
Previously we used to rely on a ledger header to match timeout event to the request, but this may not always be accurate (if two request for the same header have been sent). This changes it so that a strict (ID-based) matching is used.


siy (approved):
LGTM

siy (approved):
LGTM

talekhinezh (approved):

michalzajda (approved):

PR #247, opened 2021-04-27
Various improvements

siy:
- Implement time-based averaging for 'networkTransactionThroughput' and 'networkTransactionDemand'
- Implement error propagation from RE back to client according to specs (specs need to be updated)


radixdlt-engine/src/main/java/com/radixdlt/atom/TxErrorCode.java
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied.  See the License for the specific
+ * language governing permissions and limitations under the License.
+ */
+
+package com.radixdlt.atom;
+
+public enum TxErrorCode {
```
Sajjon, line 20: Would be great if we could catch these? Some might not be applicable...

* Amount not multiple of granularity? 
* Message longer than 255 bytes?
* Network of AccountAddress mismatches with active network? (got `"rdx1.."` but expected `"brx1.."` and vice versa)
* Network of ValidatorAddress mismatches with active network? (got `"vb1.."` but expected `"vr1.."` and vice versa)
* Network of RRI mismatches with active network? (got `".._rb1.."` but expected `".._rr1.."` and vice versa)

siy: granularity is not relevant I guess. for others I'll try to add appropriate handling

russellharvey:
Decision is not to take this PR for beta1 release; we will tackle this in beta2.

siy:
Will be replaced with better solution

Issue #251, opened 2021-05-01
CreateFixedToken default parameters values ?


cbisaillon:
When I look at com.radixdlt.atom.actions.CreateMutableToken, I see that description, iconUrl and tokenUrl have default values.
In com.radixdlt.atom.actions.CreateFixedToken, they don't have default values. Is this intentional ? If not, I can do a PR with the changes

talekhinezh:
Yup, sounds like a good idea

PR #252, opened 2021-05-02
Atom action default/null

cbisaillon:
## Description

Fixing #251 

## labels
Maintenance


talekhinezh:
@rvelaz Looks like sonarqube is failing to run on this. Can we get that enabled for accounts not part of radix team?

rvelaz:
@talekhinezh it should be good now.
@cbisaillon would you mind pushing a dummy commit to trigger the workflow?

cbisaillon:
@talekhinezh @rvelaz I made the dummy commit

PR #258, opened 2021-05-05
Refactor consensus dispatch

talekhinezh:
* Move last remaining module (Consensus) to use event dispatch method
* Cleaned up DeterministicProcessor to single class
* All Rx runners now use same construction logic


siy (approved):
LGTM with one minor comment

PR #276, opened 2021-05-12
Feature/fork view ceiling

talekhinezh:
* Add ability to change view ceiling via forks
* Add forking simulation tests
* Fix some /system endpoints


siy (approved):
LGTM

PR #283, opened 2021-05-18
Feature/proposer in vertex

talekhinezh:
* Implement corresponding actions to construction fork
* Add proposer information to vertex for use with staking and rewards

### DIFF
--- a/radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/ConstraintMachine.java
+++ b/radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/ConstraintMachine.java
@@ -139,7 +139,7 @@ public final class ConstraintMachine {
 						.filter(SystemParticle.class::isInstance)
 						.findFirst()
 						.or(() -> store.loadAddr(txn, addr))
-						.or(() -> Optional.of(new SystemParticle(0, 0, 0))); // A bit of a hack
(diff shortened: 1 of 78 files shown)


radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/ConstraintMachine.java
```diff
@@ -139,7 +139,7 @@ public ReadableAddrs immutableIndex() {
 						.filter(SystemParticle.class::isInstance)
 						.findFirst()
 						.or(() -> store.loadAddr(txn, addr))
-						.or(() -> Optional.of(new SystemParticle(0, 0, 0))); // A bit of a hack
+						.or(() -> Optional.of(new SystemParticle(0, 0, 0, null))); // A bit of a hack
```
siy, line 142: The whole supplier can be replaced with method reference and returned value - into constant. This allows to avoid unnecessary allocation and whole code would be a bit cleaner and readable.

siy (approved):
LGTM with one minor comment

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/ConstraintMachine.java
talekhinezh: Not worth making it cleaner now as this needs to be refactored towards a better architecture overall.

PR #298, opened 2021-05-26
More acceptance tests

Theoklitos:
I added 8 acceptance tests which touch most of the JSON-RPC client's methods. 

Unstaking is not tested due to betanetV1/V2 unstaking lock.


siy (approved):
LGTM

PR #301, opened 2021-05-28
Feature/refactor emissions

talekhinezh:
* Refactor emissions so that there is only a single ValidatorStake substate per validator
* Refactor unstaking so that the unstake occurs on epoch boundaries


siy (approved):
LGTM

PR #304, opened 2021-05-31
Feature/unstake xrd

talekhinezh:
* Implement client endpoints for unstaking


siy (approved):
LGTM

LukasGasior1 (approved):
LGTM

radixdlt-core/radixdlt/src/main/java/com/radixdlt/client/store/berkeley/BalanceEntry.java
```diff
@@ -57,55 +58,64 @@
...
 		@JsonProperty("owner") REAddr owner,
 		@JsonProperty("delegate") byte[] delegate,
 		@JsonProperty("rri") String rri,
 		@JsonProperty("amount") UInt384 amount,
-		@JsonProperty("negative") boolean negative
+		@JsonProperty("negative") boolean negative,
+		@JsonProperty("epochUnlocked") Long epochUnlocked
 	) {
-		return createFull(owner, delegate, rri, amount, negative);
+		return createFull(owner, delegate, rri, amount, negative, epochUnlocked);
```
iamyulong, line 74: Having the `negative` flag is adding much complexity to the clients of the class. Anyone using `getAmount()` needs to check `isNegative()` after this change. This is very error prone in my view. Can we store the effective amounts instead?

There are API handlers that format balances without interpreting `negative` flag. We need to fix them.

siy: Initial amounts are expressed using unsigned integers. As a consequence, it's impossible to use effective amounts. Also, stored values are always positive (they represent balances after all), so formatting does not need to consider the flag.

iamyulong: I think after this change, shut down balances will have negative amounts. Will API be affected in any way?

If we want to keep `UInt256` type, we change the getter method to return a `Pair<UInt256, Boolean>` to enforce clients to check at compile-time; otherwise, maybe we should consider signed integer, like `BigInteger`. What do you think?

siy: Balance can go down to zero, but should not be negative. In either case this does not affect the API.

While I like the idea of enforcing the check, I don't think it worth to apply it here. The class is rather low level internal part of `BerkeleyClientApiStore` and it's usage outside this class is very limited.

PR #308, opened 2021-06-01
Copy specific HTTP headers from request to response

siy:


michalzajda (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/RestUtils.java
```diff
@@ -117,22 +120,26 @@ public static void respond(HttpServerExchange exchange, Object object) {
...
-				.whenComplete((response, exception) -> {
-					if (exception == null) {
-						respond(exchange, response);
-					} else {
-						exchange.setStatusCode(StatusCodes.BAD_REQUEST);
-						exchange.getResponseSender().send("Unable to handle request: " + exception.getMessage());
-					}
-				});
-		});
+		exchange.dispatch(() -> CompletableFuture
```
talekhinezh, line 123: Unrelated to the change but I dont believe this respondAsync works as intended. I think undertow always expects `exchange` object to be only used in thread it is given in or in the thread supplied through `exchange.dispatch()`. Otherwise `exchange` object might change/disappear from underneath and then one gets strange race condition bugs. I ran into this issue a few times but forgot to mention it.

See `RestUtils.withBody` for how I think it should be handled.

siy: Current code uses `exchange.dispatch()` without checking the `isInIoThread` condition. This might be redundant in some cases, but otherwise should be safe in regard to preserving `exchange` during asynchronous processing.

I've updated processing to unify both processing paths (for RPC and for REST). As a side effect, this change also introduced limits on REST request body size. Previously this check was present only for RPC calls.

LukasGasior1 (approved):

PR #310, opened 2021-06-03
Transaction Logger

talekhinezh:
* Refactor how actions are computed from REProcessedTxn
* Add transaction logger for purpose of log parser


siy (approved):
LGTM

radixdlt-engine/src/main/java/com/radixdlt/atommodel/tokens/Bucket.java
```diff
@@ -16,30 +16,16 @@
...
-
-	public TxAction getTxAction() {
-		return txAction;
-	}
+// TODO: these methods are really here for client api,
+// TODO: fix to be more in line with scrypto abstractions
+public interface Bucket {
+	REAddr resourceAddr();
+	REAddr getOwner();
+	ECPublicKey getValidatorKey();
```
iamyulong, line 29: I don't know how practical it's going to be, but it would be awesome if we can introduce another address type for validator public key before mainnet. Validator should be addressable in scrypto (in future).

iamyulong (commented):
Happy to see the resource bucket structuring and action parsing refactoring. Action parsing (interpretation of the state changes) should not be part of RE validation.

radixdlt-engine/src/main/java/com/radixdlt/atommodel/tokens/Bucket.java
talekhinezh: Actually the reason why I didn't give those REAddr is that I don't think those validator keys are actually addresses. I would imagine them to be more like keys in a map maintained by the SystemStaking component

iamyulong: As discussed offline, the main reason for using single component is to maintain a global list of validators (without iterating searching the ledger state) and the top 100. Under such organisation, validator public key is just an internal key of the system staking component which exposes all staking functionality.

LukasGasior1 (approved):
LGTM

PR #312, opened 2021-06-07
Refactor fees

talekhinezh:
* Add a SYSCALL instruction which doesn't update substate but instead can update executionState
* Implemented fees using above instruction


siy (approved):
LGTM

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/ConstraintMachine.java
```diff
@@ -220,233 +168,86 @@ public SubstateCursor shutdownAll(Class<? extends Particle> particleClass) {
...
-			if (expectEnd && inst.getMicroOp() != REInstruction.REOp.END) {
-				throw new ConstraintMachineException(CMErrorCode.MISSING_PARTICLE_GROUP, validationState);
+			if (expectEnd && inst.getMicroOp() != REInstruction.REMicroOp.END) {
+				throw new ConstraintMachineException(CMErrorCode.MISSING_PARTICLE_GROUP);
 			}
 
-			if (inst.getMicroOp() == REInstruction.REOp.DOWNALL) {
+			if (inst.getMicroOp() == REInstruction.REMicroOp.SYSCALL) {
+				CallData callData = inst.getData();
+				var opSignature = OpSignature.ofMethod(inst.getMicroOp().getOp(), REAddr.ofSystem());
```
iamyulong, line 244: Do you think it's worth to make the operand of `SYSCALL` generic, such as `address + bytes`? The fee taking system contract would parse a `u256` from the `bytes`.

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/metering/FixedFeeMetering.java
```diff
@@ -0,0 +1,57 @@
...
+			var substateWithArg = (SubstateWithArg) param;
+			if (substateWithArg.getSubstate() instanceof TokensInAccount) {
+				var tokensInAccount = (TokensInAccount) substateWithArg.getSubstate();
+				if (tokensInAccount.getResourceAddr().isNativeToken()) {
+					return;
+				}
+			}
+		}
+
+		context.verifyHasReserve(fixedFee);
```
iamyulong, line 55: I understand you have a TODO here. If we call back the metering handler after each procedure call, we won't need the above condition (the first non-SYSTEM transition must be transaction fee).

iamyulong (commented):
Nice work!

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/ConstraintMachine.java
talekhinezh: Yup, the `CallData` object is supposed to be the generic operand. Avoiding use of `byte[]` so that we can possibly avoid buffer copies.

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/metering/FixedFeeMetering.java
talekhinezh: I think for fees in general we want to check there is enough fees before executing the resource consumption. No?

michalzajda (approved):

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/metering/FixedFeeMetering.java
iamyulong: Yes, ideally no state updates should be conducted before transaction fee is checked. This also avoids DoS attacks.

Please ignore my comment - The approach does not work, as we need at least two transitions to pay a fee (one DOWN and one SYSCALL).

In the meeting, we discussed the "loan" idea where a transaction is granted a loan for execution. The transaction is considered invalid if it doesn't pay back the loan before it's is drained.

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/ConstraintMachine.java
iamyulong: As discussed offline, we're not adding address ATM. Instead, we will reserve a few bytes in the calldata to distinguish different actions.

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/metering/FixedFeeMetering.java
iamyulong: The only downside is that we may need to execute the transactions in mempool, after importing a block.

Another approach is to enforce static rules - the first X instruction must be paying a fee.

We might want some approach in between.

iamyulong (approved):

PR #334, opened 2021-06-22
Add Radix transaction specs

iamyulong:
This PR adds documentation on Radix transaction specs.

It does not include the changes we've made since `2021-06-08`, for which I will make another PR.


siy (approved):
LGTM

PR #336, opened 2021-06-23
Improve error reporting for submission issues

siy:
- Search for more informative error details reported to the caller


siy:
Will be replaced with another PR

PR #340, opened 2021-06-23
Feature/track missed proposals

talekhinezh:
* Track all missed proposals in an epoch
* Verify that each epoch only has a given number of proposals


siy (approved):
LGTM

LukasGasior1 (approved):

PR #347, opened 2021-06-25
Transaction history performance improvements

siy:


radixdlt-engine/src/main/java/com/radixdlt/atommodel/system/construction/NextEpochConstructorV3.java
```diff
@@ -251,6 +254,6 @@ public void construct(SystemNextEpoch action, TxBuilder txBuilder) throws TxBuil
 		validatorKeys.forEach(k -> txBuilder.up(new ValidatorBFTData(k, 0)));
 
 		txBuilder.up(new EpochData(prevEpoch.getEpoch() + 1));
-		txBuilder.up(new RoundData(0, 0));
+		txBuilder.up(new RoundData(0, timestamp));
```
iamyulong, line 257: Shouldn't this `timestamp` be provided by the action?

siy: I don't think so. This is a "system" time, user has no control over it.

LukasGasior1 (approved):

PR #354, opened 2021-06-27
Feature/remove invalid txns

talekhinezh:
* Fix issue where leaders are proposing invalid txns due to mempool not evicting correct txns


radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/REProcessedTxn.java
```diff
@@ -80,6 +83,23 @@ public boolean isSystemOnly() {
...
+			.flatMap(i -> {
+				if (i.getMicroOp() == REInstruction.REMicroOp.DOWN || i.getMicroOp() == REInstruction.REMicroOp.READ) {
+					SubstateId substateId = i.getData();
+					return Stream.of(substateId);
+				} else if (i.getMicroOp() == REInstruction.REMicroOp.VDOWN || i.getMicroOp() == REInstruction.REMicroOp.VREAD) {
+					Substate substate = i.getData();
+					return Stream.of(substate.getId());
+				} else if (i.getMicroOp() == REInstruction.REMicroOp.VDOWNARG) {
+					Pair<Substate, byte[]> substateArg = i.getData();
+					return Stream.of(substateArg.getFirst().getId());
```
iamyulong, line 97: Do we need to consider `DOWNALL` and `DOWNINDEX`?

iamyulong (approved):
Left a question, otherwise looks good to me.

siy (approved):
LGTM

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/REProcessedTxn.java
talekhinezh: No, in those cases, transactions dependent on those substates should not be affected whether they are still valid, mostly because the only transactions using those are sys transactions

PR #358, opened 2021-06-27
Feature/cleanups plus nodes keys

rvelaz:
key generation for fullnodes


siy (approved):
LGTM

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/PeerChannel.java
```diff
@@ -205,6 +206,16 @@ public void channelRead0(ChannelHandlerContext ctx, byte[] buf) throws Exception
 
 	@Override
 	public void channelInactive(ChannelHandlerContext ctx) {
+		final ValidatorAddressing validatorAddresses = ValidatorAddressing.bech32("vb");
```
LukasGasior1, line 209: this shouldn't be hardcoded to "vb", better to use injected `Addressing`, also I think what's meant here is `NodeAddressing`, not validator (if it's meant to be a validator addr, then we should change the log message)

LukasGasior1 (approved):

PR #360, opened 2021-06-28
Port validator list retrieval and transaction history performance fixes

siy:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/data/ValidatorInfoDetails.java
```diff
@@ -96,18 +107,52 @@ public ECPublicKey getValidatorKey() {
...
 		return jsonObject()
 			.put("address", getValidatorAddress())
 			.put("ownerAddress", AccountAddress.of(owner))
-			.put("name", name)
-			.put("infoURL", infoUrl)
+			.putOpt("name", name)
+			.putOpt("infoURL", infoUrl)
 			.put("totalDelegatedStake", totalStake)
 			.put("ownerDelegation", ownerStake)
+			.put("percentage", percentage)
```
iamyulong, line 154: Suggest renaming to `rakePercentage`

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/AccountInfoService.java
```diff
@@ -56,16 +58,35 @@ public JSONObject getAccountInfo() {
...
+				.put("registered", details.isRegistered())
+				.put("owner", AccountAddress.of(details.getOwner()))
+				.put("percentage", details.getPercentage())
+				.put("allowDelegation", details.isExternalStakesAllowed()),
+			() -> result
+				.put("name", "")
+				.put("url", "")
+				.put("registered", false)
+				.put("owner", AccountAddress.of(REAddr.ofPubKeyAccount(bftKey)))
+				.put("percentage", 0)
```
iamyulong, line 85: Again

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/SubmissionService.java
```diff
@@ -115,14 +118,38 @@ private TxnConstructionRequest toConstructionRequest(
...
+
+		while (reportedException.getCause() instanceof ConstraintMachineException) {
+			reportedException = reportedException.getCause();
+		}
+
+		if (reportedException instanceof  ConstraintMachineException && reportedException.getCause() != null) {
+			reportedException = reportedException.getCause();
+		}
+
+		return reportedException;
```
iamyulong, line 150: Can we report the full exception stack to end user?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/reducer/AllValidators.java
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	public UInt256 getStake(ECPublicKey validatorKey) {
+		return ofNullable(stake.get(validatorKey))
+			.orElse(UInt256.ZERO);
+	}
+
+	public Boolean allowsDelegation(ECPublicKey validatorKey) {
+		return ofNullable(delegationFlags.get(validatorKey))
+			.orElse(Boolean.TRUE);
```
iamyulong, line 139: I was wondering if it would make more sense to return an `Optional` instead provides a default. Also I thought we changed the delegation flag default to `false`, didn't we?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/reducer/AllValidators.java
```diff
@@ -0,0 +1,265 @@
...
+
+import static com.radixdlt.utils.functional.FunctionalUtils.removeKey;
+import static com.radixdlt.utils.functional.FunctionalUtils.replaceEntry;
+
+import static java.util.Optional.ofNullable;
+
+/**
+ * Wrapper class for registered validators
+ */
+public final class AllValidators {
```
iamyulong, line 43: Not directly related to your CR but a general question.

I've noticed that many of our classes are modelled as immutable. Do we have general guideline about which class should be immutable?

It seems to me that we're making a lot updates to `AllValidator`?  Is it worth to sacrifice performance  (a lot of copies) for immutability.

cc @talekhinezh

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/SubmissionService.java
talekhinezh: Agree, this would probably be most helpful

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/reducer/AllValidators.java
talekhinezh: Yup, it should be defaulted to false

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/reducer/AllValidators.java
talekhinezh: Good question. No, I don't think we should stick to immutability for the sake of immutability. Some of these reducer classes are set as immutable because of the way we manage these in our state branching. Changing state and then possibly needing to revert that state was faster to implement with immutable structures. Having said that, I'm definitely looking to removing the immutable reducer stuff soon.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/SubmissionService.java
siy: I disagree. User wants to know the real cause rather than dig through full stack trace (which is quite long in this case). The full stack trace is provided in the logs.

iamyulong: I think it depends on what you want to achieve: 
1. To provide failure cause to the end users so that they know what to do next;
2. To provide a trace of failures, so that app developers can diagnose the problem without consulting the node.

For 1, we need to provide explicit list of error codes with documentation on the suggested solution.
For 2, it's better to provide as much details as possible IMHO.

What do you think?

siy: It's a user-facing API (for example, wallet uses it), so the aim to achieve #1. Unfortunately, at the point where I'm intercepting the exception, I have no detailed information such as error code. Actually, the whole method is an attempt to retrieve some human-recognizable information instead of just the transaction blob which we did return previously.
Also, it was unexpected, but I've discovered that messages returned by this method are much more helpful even for the developer: simple text search with the error cause immediately points to the correct location in code.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/reducer/AllValidators.java
siy: Fixed

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/reducer/AllValidators.java
siy: I'd prefer to use immutable structures unless I have sensible reasons to have value mutable. Mutable state requires considering possible concurrent execution issues, which creates unnecessary mental overhead.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/data/ValidatorInfoDetails.java
siy: Fixed

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/AccountInfoService.java
siy: Fixed

LukasGasior1 (approved):

iamyulong (approved):

PR #361, opened 2021-06-28
Update transaction specs

iamyulong:
Update transaction specs to include our latest changes:
- Substate index definition change
- Tokens and PrepareStake substate schema change
- HRP change (missed last time)
- New fee payment
- New token creation mechanism


Sajjon:
@iamyulong I propose changing the Debug output of bytes to this:

Screenshot 2021-06-28 at 20 52 58

Which gives output:
```sh
|- SYSCALL(0x000000000000000000000000000000000000000000000000000de0b6b3a7640000)
```

Instead of harder to read:
```sh
|- SYSCALL(Bytes { length: 33, data: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 13, 224, 182, 179, 167, 100, 0, 0] })
```

Sajjon:
So `SYSCALL` should either be:
1. `0x00 || UInt256` 
2. or: `0x01 || UInt256` 

Both Having length 33, excluding INS OPCODE `0x09`.

So `FEE_RESERVE_PUT` would be:
`092100<32 more bytes>`

and a `FEE_RESERVE_TAKE` would be:
`092101<32 more bytes>`

?

iamyulong:
@Sajjon I will update `Bytes` debug format as suggested

And yes, your understanding about `SYSCALL` is correct. The `calldata` is wrapped into a `Bytes` and is of length `33` for both cases.

siy (approved):
LGTM

PR #362, opened 2021-06-29
Feature/manage int256 overflows

talekhinezh:
* Cleanup TokenResource substate
* Add scratchpad for updating validators (rather than using immutable structures)
* Add some overflow checks


siy (approved):
LGTM

michalzajda (approved):

PR #369, opened 2021-06-30
Add more test vectors

iamyulong:
Add more test vectors


iamyulong:
Cancelling... Will combine this with our latest transaction format change.

PR #376, opened 2021-07-01
Feature/single parameter submit

siy:
Changes summary:
- Reworked parameters for finalize transaction and submit transaction
- Added support for different networks to client library
- Fixed tests
- Fixed broken notifications to client store


radixdlt-java/radixdlt-java/src/main/java/com/radixdlt/client/lib/api/token/Amount.java
```diff
@@ -0,0 +1,51 @@
...
+	static Amount amount(UInt256 value) {
+		return () -> value;
+	}
+
+	static Amount amount(long value) {
+		return () -> UInt256.from(value);
+	}
+
+	default UInt256 tokens() {
+		return subunits().multiply(UInt256.TEN.pow(SUB_UNITS_POW_10));
```
michalzajda, line 35: I have a feeling that we could use variable 18_ZEROS across the code base

radixdlt-core/radixdlt/src/main/java/com/radixdlt/NetworkGenesis.java
```diff
@@ -0,0 +1,32 @@
...
+import com.radixdlt.atom.Txn;
+import com.radixdlt.networks.Network;
+
+import java.util.Optional;
+
+public final class NetworkGenesis {
+	private NetworkGenesis() { }
+
+	//Moved from Network
+	public static Optional<Txn> txn(Network network) {
```
talekhinezh, line 29: There should be no layer of indirection between Network and hardcoded genesis, so Genesis Txn should be in Network class.

talekhinezh: If this is a problem because radixdlt-common doesn't have `Txn` then we can just change the type to be a Hex string.

radixdlt-java-common/src/main/java/com/radixdlt/identifiers/Naming.java
```diff
@@ -0,0 +1,27 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied.  See the License for the specific
+ * language governing permissions and limitations under the License.
+ */
+
+package com.radixdlt.identifiers;
+
+import java.util.regex.Pattern;
+
+public final class Naming {
```
talekhinezh, line 22: I'm not sure about moving this logic outside of `CMAtomOS`. It's the OS which makes the rules and not a "common" rule. If this needs to change this would be part of something like the `ForkConfiguration` so would rather have this be closer to RadixEngine.

radixdlt-java-common/src/main/java/com/radixdlt/networks/Network.java
```diff
@@ -61,12 +59,8 @@ public int getId() {
 		return id;
 	}
 
-	public Optional<Txn> genesisTxn() {
```
talekhinezh, line 64: Like mentioned above this should remain here. Perhaps just change it to a String.

radixdlt-java/radixdlt-java/src/main/java/com/radixdlt/client/lib/api/token/Amount.java
```diff
@@ -0,0 +1,51 @@
...
+ * either express or implied.  See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ */
+
+package com.radixdlt.client.lib.api.token;
+
+import com.radixdlt.utils.UInt256;
+
+public interface Amount {
```
talekhinezh, line 23: I think there's a lot of issues with this interfaces. Representing all forms of subunits in the same `UInt256` I think makes things confusing. Also there's no overflow or underflow protection. Would rather use something more similar to what is currently `com.radixdlt.application.tokens.Amount.java`

radixdlt-java/radixdlt-java/src/test/java/com/radixdlt/client/lib/api/sync/SyncRadixApiCreationTest.java
```diff
@@ -149,14 +160,11 @@ public void testBuildAndSubmitTransactionWithMessage() {
...
-				.onSuccess(finalizedTransaction -> client.transaction().finalize(finalizedTransaction)
+				.onSuccess(finalizedTransaction -> client.transaction().finalize(finalizedTransaction, false)
 					.onSuccess(txDTO -> assertNotNull(txDTO.getTxId()))
-					.map(txDTO -> finalizedTransaction.withTxId(txDTO.getTxId()))
 					.onSuccess(submittableTransaction -> client.transaction().submit(submittableTransaction)
 						.onFailure(failure -> fail(failure.toString()))
-						.onSuccess(txDTO -> submittableTransaction.rawTxId()
-							.ifPresentOrElse(aid -> assertEquals(aid, txDTO.getTxId()), () -> fail("Should not happen")))))
-			);
+						.onSuccess(txDTO -> assertEquals(submittableTransaction.getTxId(), txDTO.getTxId())))));
```
talekhinezh, line 167: Seems like this is turning into callback hell? I question how hard it would be to debug if an exception were to occur?

talekhinezh (changes requested):

talekhinezh (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/NetworkGenesis.java
siy: Yep, the reason was that I did need the `Txn` in the `radixdlt-java-common` to keep this method inside `Network`. 
Moved it back to `Network` and changed return type to hex string.

radixdlt-java-common/src/main/java/com/radixdlt/networks/Network.java
siy: Fixed

radixdlt-java/radixdlt-java/src/main/java/com/radixdlt/client/lib/api/token/Amount.java
siy: The idea behind this interface is to make code readable closer to plain English: `amount(10).tokens()` or `amount(100).nanos()`. It is also closer to how we usually express the amounts in real life - "10 bucks" or "15 cents". 
Since this interface is part of the client library, I feel that it will be more convenient for end-users.

Since there is only a constructor from `long`, I doubt that overflow/underflow is an issue.

radixdlt-java/radixdlt-java/src/test/java/com/radixdlt/client/lib/api/sync/SyncRadixApiCreationTest.java
siy: The client code never throws an exception (this is the exact purpose of using `Result<T>` as a value container). Any exception which may occur can be only in the user code, so it will be straightforward to nail it down.
Also, given the complexity of the code (three steps of transaction submission, where each step may fail + assertions necessary only for test), it is rather clean and far from being callback hell. In real use cases this code could be split into few smaller functions, like, for example, this is done in `SyncRadixApiTest` (`makeStake()` + `makeStake(RadixApi, UInt256)`, for example).

PR #377, opened 2021-07-01
Feature/read index

talekhinezh:


siy (approved):
LGTM

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/EpochProofVerifierV2.java
```diff
@@ -18,83 +18,62 @@
...
-			// Verify that output of radix engine and signed output match
-			// TODO: Always follow radix engine as its a deeper source of truth and just mark validator
-			// TODO: set as malicious (RPNV1-633)
-			if (epochChangeFlag) {
-				if (metadata == null) {
-					throw new IllegalStateException();
+			if (!nextEpochEvents.isEmpty()) {
+				// TODO: Move this check into Meter
+				if (i != txns.size() - 1) {
+					throw new MetadataException("Additional txns added to end of epoch.");
```
iamyulong, line 47: Do you think it's worth to have something like block-level validation?
- The max signature per block
- The max number of epoch update per block
- Epoch update should be the last state update

Moving everything to meters doesn't sound like the right direction

radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/berkeley/BerkeleyLedgerEntryStore.java
```diff
@@ -363,7 +365,7 @@ private void open() {
 							} else if (substateTypeId == SubstateTypeId.PREPARED_RAKE_UPDATE.id()) {
 								prefixIndexSize = 2 + Long.BYTES;
 							} else if (substateTypeId == SubstateTypeId.VALIDATOR_STAKE_DATA.id()) {
-								prefixIndexSize = 3;
+								prefixIndexSize = 3 + UInt256.BYTES + ECPublicKey.COMPRESSED_BYTES;
```
iamyulong, line 368: Are we missing the `is_registered` byte?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/berkeley/BerkeleyLedgerEntryStore.java
```diff
@@ -524,13 +527,26 @@ private static OptionalLong headerKeyEpoch(DatabaseEntry entry) {
...
 		private void open() {
 			this.cursor = db.openCursor(dbTxn, null);
-			this.status = cursor.getSearchKeyRange(key, substateIdBytes, value, null);
+			if (reverse) {
+				var copy = Arrays.copyOf(indexableBytes, indexableBytes.length);
+				var lastIndex = indexableBytes.length - 1;
+				if (copy[lastIndex] == (byte) (0xff)) {
+					throw new IllegalStateException("Unexpected last byte.");
+				}
+				copy[lastIndex]++;
```
iamyulong, line 542: This doesn't look right to me.
- If the last byte is `0xff`, we should extend the indexable byte by 1 more byte
- What if there are duplicates? (although it's unlikely for validator stake data)

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/ConstraintMachine.java
```diff
@@ -387,7 +389,14 @@ public Particle next() {
...
 				} else if (inst.getMicroOp() == REInstruction.REMicroOp.SIG) {
-					metering.onSigInstruction(context);
+					if (context.permissionLevel() != PermissionLevel.SYSTEM) {
+						metering.onSigInstruction(context);
+					}
+				} else {
+					// Collect no-ops here
+					if (inst.getMicroOp() != REInstruction.REMicroOp.MSG) {
+						throw new ProcedureException("Unknown op " + inst.getMicroOp());
+					}
```
iamyulong, line 399: We are missing `HEADER`

talekhinezh: oof, good catch, need to add some tests with `HEADER`

radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/berkeley/BerkeleyLedgerEntryStore.java
talekhinezh: This is why the check at line 539 doesnt allow `0xff`.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/berkeley/BerkeleyLedgerEntryStore.java
talekhinezh: '3' used instead of '2'. I'll add more comments.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/EpochProofVerifierV2.java
talekhinezh: We don't have a concept of blocks though. We will need metering in the long run though to have limits on computation per "block".

I don't think the idea of metering against a validatorSet is a bad concept as it reinforces the idea that we must protect against malicious quorums/cartels.

iamyulong (approved):

PR #390, opened 2021-07-07
Client library docs and other improvements

siy:
Changes summary:
- Added documentation to all versions of java client 
- Fixed implementation of some network API's in client
- Fixed parameters for create fixed supply token action
- Renamed `from` to `to` in the create fixed supply token action


michalzajda (approved):

Theoklitos (approved):

iamyulong (commented):
LGTM.

Question: can we generate the DTO classes from our JSON-RPC specs?

PR #392, opened 2021-07-07
Fix acceptance tests

Theoklitos:
Updated the broken acceptance tests to use the new JSON-RPC API (and pass). Also, switched to using the Imperative API.

Several tests are missing because fork overrides are still needed. I basically need shorter epochs. This will be done in a future PR, along with the missing tests.


radixdlt-java/radixdlt-java/src/main/java/com/radixdlt/client/lib/dto/Channel.java
```diff
@@ -0,0 +1,76 @@
+/*
```
siy, line 1: Please, remove

radixdlt-java/radixdlt-java/src/main/java/com/radixdlt/client/lib/dto/Peer.java
```diff
@@ -0,0 +1,78 @@
+/*
```
siy, line 1: Please, remove

siy (approved):
LGTM with few comments

PR #409, opened 2021-07-16
Feature/add fees

talekhinezh:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/handler/DeveloperHandler.java
```diff
@@ -128,14 +131,28 @@ public JSONObject handleParseTxn(JSONObject request) {
...
+							break;
+						case "node":
+							pair = NodeAddressing.parseUnknownHrp(address);
+							break;
+						case "validator":
+							pair = ValidatorAddressing.parseUnknownHrp(address);
+							break;
+						default:
+							throw new IllegalArgumentException("type must be: [account|node|validator]");
+					}
```
siy, line 153: I think code will be much more readable if `switch` will be extracted into dedicated method.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/handler/DeveloperHandler.java
```diff
@@ -128,14 +131,28 @@ public JSONObject handleParseTxn(JSONObject request) {
...
+					var hrp = pair.getFirst();
+					var pubKey = pair.getSecond();
 					return jsonObject()
 						.put("hrp", hrp)
 						.put("public_key", Bytes.toHexString(pubKey.getCompressedBytes()));
```
siy, line 158: Alternative version:
```
return pair.map((hrp, pubKey) -> 
    jsonObject()
        .put("hrp", hrp)
	.put("public_key", Bytes.toHexString(pubKey.getCompressedBytes())));
```

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/handler/DeveloperHandler.java
```diff
@@ -147,14 +164,29 @@ public JSONObject handleParseAddress(JSONObject request) {
...
+							break;
+						case "node":
+							address = NodeAddressing.bech32(hrp).of(publicKey);
+							break;
+						case "validator":
+							address = ValidatorAddressing.bech32(hrp).of(publicKey);
+							break;
+						default:
+							throw new IllegalArgumentException("type must be: [account|node|validator]");
+					}
```
siy, line 188: I think it makes sense to extract this logic into dedicated method.

siy (approved):
LGTM with comments

iamyulong (approved):
Looks good to me, but I have one question regarding encoding of optional fields:
- Why do we need to encode a zero-value for nones? Is it to make substates of fixed length for better indexing?
- I was also thinking maybe we should index based on fields (rather than prefix) for flexibility.

PR #416, opened 2021-07-18
Don't allow trailing bytes in syscall calldata

talekhinezh:


siy (approved):
LGTM

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/CallData.java
```diff
@@ -38,10 +39,15 @@ public byte get(int offset) throws CallDataAccessException {
...
 		if (offset < 0 || (offset + UInt256.BYTES) > data.length) {
 			throw new CallDataAccessException(data.length, offset, UInt256.BYTES);
 		}
+
+		if (data.length > offset + UInt256.BYTES) {
+			throw new TrailingBytesException("Call data has " + data.length + " bytes.");
+		}
+
 		return UInt256.from(data, offset);
 	}
```
iamyulong, line 52: Thanks for making this PR! I should've done this. 

Given that all our system functions accept a single paremters, is it worth to refactor this class as follows? 
```
private byte method;
private byte[] args;

public CallData(byte[] data) {
        Preconditions.checkArgument(data != null && data.length >= 1);
        this.method = data[0];
        this.args = Arrays.copyOfRange(data, 1, data.length);
}

public Syscall getMethod() {

}

public UInt256 getArgsAsUint256() {

}

public byte[]  getArgsAsBytes() {

}
```

iamyulong: Also, I was thinking if we should encode the length of `rri` in the calldata, as it's variable length bytes. Now, we depend on the EOF.

talekhinezh: Is there a clear advantage in relying on an EOF?

PR #418, opened 2021-07-18
Add ValidatorSystemMetadata

talekhinezh:


siy:
What is the purpose of `UpdateValidatorSystemMetadata`? Is it a regular action which can be issued via API?

iamyulong (approved):
This is an interesting PR. 

~~As Sergiy asked, what's the intended use of this 32-byte data for validators?~~   (It's used for storing auto-update voting info)

siy (approved):

PR #437, opened 2021-07-21
Blacklist uri on failed p2p handshake

LukasGasior1:
This PR adds blacklisting of a single node URI, in contrast to banning a peer (all known URIs for a node id).
The new mechanism is used to blacklist an URI when a p2p handshake fails.
Handshake response packet now includes an extra prefix byte indicating the status (0x01 = success, 0x02 = error) which is used to signal a failure while handling auth initiate message back to the initiator.

### DIFF
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/SystemConfigService.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/SystemConfigService.java
@@ -463,7 +463,9 @@ public class SystemConfigService {
 		final var knownAddressesArray = jsonArray();
 
 		e.getKnownAddresses().forEach(addr -> {
-			final var addrObj = jsonObject().put("uri", addr.getUri());
+			final var addrObj = jsonObject()
+				.put("uri", addr.getUri())
+				.put("blacklisted", addr.blacklisted());
 			addr.getLastSuccessfulConnection().ifPresent(ts -> addrObj.put("lastSuccessfulConnection", ts));
 			knownAddressesArray.put(addrObj);
 		});
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
@@ -245,6 +245,13 @@ public final class AddressBookEntry {
 		}
 	}
 
+	public AddressBookEntry cleanupExpiredBlacklsitedUris() {
+		final var newKnownAddresses = knownAddresses.stream()
+			.filter(not(PeerAddressEntry::blacklistExpired))
+			.collect(ImmutableSet.toImmutableSet());
+		return new AddressBookEntry(nodeId, bannedUntil, newKnownAddresses);
+	}
+
 	@Override
 	public String toString() {
 		return String.format(
@@ -314,6 +321,10 @@ public final class AddressBookEntry {
 
 		public boolean blacklisted() {
 			return blacklistedUntil.filter(v -> v.isAfter(Instant.now())).isPresent();
+		}
+
+		public boolean blacklistExpired() {
+			return blacklistedUntil.isPresent() && !blacklisted();
 		}
 
 		@JsonProperty("lastSuccessfulConnection")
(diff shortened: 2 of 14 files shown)


radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/SystemConfigService.java
```diff
@@ -463,7 +463,9 @@ private JSONObject addressBookEntryToJson(AddressBookEntry e) {
 		final var knownAddressesArray = jsonArray();
 
 		e.getKnownAddresses().forEach(addr -> {
-			final var addrObj = jsonObject().put("uri", addr.getUri());
+			final var addrObj = jsonObject()
+				.put("uri", addr.getUri())
+				.put("blacklisted", addr.blacklisted());
```
siy, line 468: Please, update https://github.com/radixdlt/open-rpc-spec accordingly.

siy (approved):
LGTM with comments

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/handshake/AuthHandshaker.java
```diff
@@ -164,57 +164,64 @@ private AuthInitiateMessage createAuthInitiateMessage(ECPublicKey remotePubKey)
...
+			this.responsePacketOpt = Optional.of(packet);
+			this.remotePubKeyOpt = Optional.of(remotePubKey);
 
-		final var handshakeResult = finalizeHandshake(remoteEphemeralKey, message.getNonce());
-		return Pair.of(packet, handshakeResult);
+			final var handshakeResult = finalizeHandshake(remoteEphemeralKey, message.getNonce());
+			return Pair.of(packet, handshakeResult);
+		} catch (PublicKeyException | InvalidCipherTextException ex) {
+			return Pair.of(null, AuthHandshakeResult.error("Handshake decryption failed", Optional.empty()));
+		}
```
iamyulong, line 210: It looks odd to return a `Result` while throwing `IOException`: callers need to consider both returns and exceptions.

A more elegant way of handling this would be to catch all `RuntimeException` and checked exceptions, and return an `Error`; otherwise, returns a `Success` result.

michalzajda (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
```diff
@@ -202,7 +210,33 @@ public AddressBookEntry withLastSuccessfulConnectionFor(RadixNodeUri uri, Instan
...
+			final var newAddressEntry = new PeerAddressEntry(uri, Optional.of(lastSuccessfulConnectionFor), false);
+			final var newKnownAddresses = ImmutableSet.<PeerAddressEntry>builder()
+				.addAll(this.knownAddresses)
+				.add(newAddressEntry)
+				.build();
+			return new AddressBookEntry(nodeId, bannedUntil, newKnownAddresses);
+		}
+	}
+
+	public AddressBookEntry withBlacklistedUri(RadixNodeUri uri) {
```
iamyulong, line 222: Do we have a retention policy for this growing URI deny list?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/handshake/AuthHandshaker.java
LukasGasior1: yeah, I think it makes sense to catch IOException as well cause that's thrown by dson deserialization and indicates an invalid initiate msg. I wouldn't catch RuntimeExceptions though as they might be caused by anything really, not necessarily a faulty handshake message.
Alternatively we could model different errors with enums, but I think I'd rather just let RuntimeExceptions be handled by the upper layer.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
LukasGasior1: Good point. I think we'd generally want to add a limit on address book entries stored.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/handshake/AuthHandshaker.java
iamyulong: Sounds good!

iamyulong (approved):

PR #443, opened 2021-07-22
Fix p2p auth handshake backwards compatibility

LukasGasior1:
This brings back p2p backwards compatibility.
Instead of signaling auth handshake error back to the "client" explicitly with a status byte, client side now assumes that if connection is closed at the auth handshake state then it's the handshake that failed. It is possible that the connection is closed for other reason, unrelated to handshake, but it's rather unlikely and even if so, it won't do any damage (other than temporarily blacklisting an URI).

### DIFF
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PeerEvent.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PeerEvent.java
@@ -66,8 +66,6 @@ package com.radixdlt.network.p2p;
 
 import com.radixdlt.network.p2p.transport.PeerChannel;
 
-import java.util.Objects;
-
 public interface PeerEvent {
 
 	final class PeerConnected implements PeerEvent {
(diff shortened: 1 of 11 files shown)


radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PeerEvent.java
```diff
@@ -85,23 +83,6 @@ private PeerConnected(PeerChannel channel) {
...
-		@Override
-		public boolean equals(Object o) {
-			if (this == o) {
-				return true;
-			}
-			if (o == null || getClass() != o.getClass()) {
-				return false;
-			}
-			final var that = (PeerConnected) o;
-			return Objects.equals(channel, that.channel);
```
LukasGasior1, line 98: equals/hashCodes using the channel didn't really make sense in these events classes

siy (approved):
LGTM

michalzajda (approved):

talekhinezh (approved):

PR #446, opened 2021-07-23
Feature/token def into genesis builder

talekhinezh:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/service/ActionParserService.java
```diff
@@ -227,6 +227,11 @@ public ActionParserService(Addressing addressing) {
 		}
 	}
 
+	private static Result<ECPublicKey> pubKeyOrSystem(JSONObject element) {
```
siy, line 230: I think it's better to return `Result<Optional>` here instead of `Result` with `null` inside.

siy (approved):
LGTM with one comment

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/checkpoint/GenesisProvider.java
```diff
@@ -128,6 +130,14 @@ public VerifiedTxnsAndProof get() {
 			.collect(ImmutableMap.toImmutableMap(TokenIssuance::receiver, TokenIssuance::amount, UInt256::add));
 
 		var actions = new ArrayList<TxAction>();
+		actions.add(new CreateMutableToken(
+			null,
+			"xrd",
+			"Rads",
+			"Radix Tokens",
+			"",
+			""
```
iamyulong, line 139: Do we need to update the genesis transaction because of the URL changes?

talekhinezh: GenesisProvider is just used for testnets

LukasGasior1 (approved):

iamyulong (approved):

PR #456, opened 2021-07-26
don't emit handshake failed event in channelInactive while in AUTH_HA…

LukasGasior1:
Don't emit handshake failed event in channelInactive while in AUTH_HANDSHAKE state


iamyulong (approved):
LGTM

talekhinezh (approved):

PR #489, opened 2021-09-01
Bring back "docker" tests

Theoklitos:
Re-implemented the old "docker" tests.

### Description

Contains the code used to start docker networks of radix nodes via java (using [docker-client](https://github.com/docker-java/docker-java)) plus some fixtures to help with running tests.

[Check out the project's README](https://github.com/radixdlt/radixdlt/tree/feature/RPNV1-1422-system-tests/radixdlt-regression/system-tests) and try to run these tests locally with:
```
./gradlew systemTests
```

There is only one such "system test" included in this PR. [See it here](https://github.com/radixdlt/radixdlt/blob/feature/RPNV1-1422-system-tests/radixdlt-regression/system-tests/src/test/java/com/radixdlt/test/system/SmokeTesting.java). It does the following:
1. starts a network with 3 nodes
2. asserts liveness (in a crude fashion)
3. Restarts a node
4. Makes sure that the restarted node eventually gets UP and that the network has liveness (waiting a minute at most).

I will enable system testing in Jenkins in a subsequent PR.


radixdlt-regression/src/main/java/com/radixdlt/test/network/checks/Checks.java
```diff
@@ -0,0 +1,53 @@
...
+        checks.put(LivenessCheck.class, livenessCheck);
+    }
+
+    public boolean runCheck(String name, Object... options) {
+        AtomicReference<Check> checkWithMatchingName = new AtomicReference<>();
+        checks.forEach((key, value) -> {
+            if (key.getSimpleName().toLowerCase().contains(name.toLowerCase())) {
+                if (checkWithMatchingName.get() == null) {
+                    checkWithMatchingName.set(value);
+                }
```
siy, line 37: Code above can be simplified to `checkWithMatchingName.compareAndSet(null, value)`. Value will be set only once.

radixdlt-regression/src/main/java/com/radixdlt/test/utils/universe/UniverseUtils.java
```diff
@@ -0,0 +1,73 @@
...
+
+    }
+
+    /**
+     * is probably not thread safe
+     */
+    public static UniverseVariables generateEnvironmentVariables(int numberOfValidators) {
+        // create a custom stream to hold system.out output until the generation finishes.
+        // this could be super messy when run concurrently
+        String outputAsString = getUniverseGenerationOutputAsString(numberOfValidators);
```
siy, line 27: can be `var` (several lines)

siy (approved):
Few minor comments, otherwise LGTM.
File `default.config` probably should be removed.

PR #491, opened 2021-09-08
Feature/fix user tx history

talekhinezh:
* Synchronize the user transaction history with the ledger itself
* Remove use of timestamps as cursor

### DIFF
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/accounts/BerkeleyAccountTxHistoryStore.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/accounts/BerkeleyAccountTxHistoryStore.java
@@ -208,17 +208,19 @@ public class BerkeleyAccountTxHistoryStore implements BerkeleyAdditionalStore {
 		}
 	}
 
+	private void storeTxnForAccount(Transaction dbTxn, REProcessedTxn txn, REAddr addr) {
+		var offset = nextOffset(dbTxn, addr);
+		var key = key(addr, offset);
+		var value = new DatabaseEntry(txn.getTxnId().getBytes());
+		var status = accountTxHistory.putNoOverwrite(dbTxn, key, value);
+		if (status != OperationStatus.SUCCESS) {
+			throw new IllegalStateException("Unable to store account transaction(off: " + offset + "): " + status);
+		}
+	}
+
 	@Override
 	public void process(Transaction dbTxn, REProcessedTxn txn, long stateVersion, Function<SystemMapKey, Optional<RawSubstateBytes>> mapper) {
 		getAccountsAssociatedWithTxn(txn)
-			.forEach(addr -> {
-				var offset = nextOffset(dbTxn, addr);
-				var key = key(addr, offset);
-				var value = new DatabaseEntry(txn.getTxnId().getBytes());
-				var status = accountTxHistory.putNoOverwrite(dbTxn, key, value);
-				if (status != OperationStatus.SUCCESS) {
-					throw new IllegalStateException("Unable to store account transaction(off: " + offset + "): " + status);
-				}
-			});
+			.forEach(addr -> storeTxnForAccount(dbTxn, txn, addr));
 	}
 }
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/accounts/BerkeleyAccountTxHistoryStore.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/accounts/BerkeleyAccountTxHistoryStore.java
@@ -1,0 +1,224 @@
+/*
+ * Copyright 2021 Radix DLT Ltd incorporated in England.
+ * Licensed under the Radix License, Version 1.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at:
+ *
+ * radixfoundation.org/licenses/LICENSE-v1
+ * The Licensor hereby grants permission for the Canonical version of the Work to be
+ * published, distributed and used under or by reference to the Licensor’s trademark
+ * Radix ® and use of any unregistered trade names, logos or get-up.
+ *
+ * The Licensor provides the Work (and each Contributor provides its Contributions) on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied,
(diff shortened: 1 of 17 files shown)


radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/accounts/BerkeleyAccountTxHistoryStore.java
```diff
@@ -0,0 +1,224 @@
...
+				if (status != SUCCESS) {
+					throw new NoSuchElementException();
+				}
+				var nextOffset = Longs.fromByteArray(key.getData(), REAddr.PUB_KEY_BYTES);
+				var nextValue = AID.from(value.getData());
+				var curStatus = cursor.getPrev(key, value, null);
+				status = curStatus == SUCCESS && keyIsREAddr(key, addr) ? SUCCESS : NOTFOUND;
+				return Pair.of(nextValue, nextOffset);
+			}
+		};
```
diegomrsantos, line 159: I'd extract the anonymous class declaration to a new descriptive name. The base of my suggestion is the chapter about functions in the clean code book. Here is a summary https://www.fabrizioduroni.it/2019/07/28/clean-code-functions/.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/accounts/BerkeleyAccountTxHistoryStore.java
```diff
@@ -0,0 +1,224 @@
...
+			.forEach(addr -> {
+				var offset = nextOffset(dbTxn, addr);
+				var key = key(addr, offset);
+				var value = new DatabaseEntry(txn.getTxnId().getBytes());
+				var status = accountTxHistory.putNoOverwrite(dbTxn, key, value);
+				if (status != OperationStatus.SUCCESS) {
+					throw new IllegalStateException("Unable to store account transaction(off: " + offset + "): " + status);
+				}
+			});
+	}
```
diegomrsantos, line 223: I think I'd extract it to a method as well.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/module/ArchiveEndpointModule.java
```diff
@@ -125,6 +125,13 @@ public JsonRpcHandler accountGetUnstakePositions(ArchiveAccountHandler archiveAc
 		return archiveAccountHandler::handleAccountGetUnstakePositions;
 	}
 
+	@ArchiveEndpoint
+	@ProvidesIntoMap
+	@StringMapKey("account.get_transaction_history2")
```
diegomrsantos, line 130: Can we write a more descriptive name describing the difference between this method and the existing `account.get_transaction_history`?

radixdlt-java/radixdlt-java/src/main/java/com/radixdlt/client/lib/api/async/AsyncRadixApi.java
```diff
@@ -268,6 +271,16 @@
 			return call(request, new TypeReference<>() {});
 		}
 
+		@Override
+		public Promise<TransactionHistory2> history2(
```
diegomrsantos, line 275: Can we write a more descriptive name describing the difference between this method and the existing `history`?

radixdlt-java/radixdlt-java/src/main/java/com/radixdlt/client/lib/dto/Transaction2DTO.java
```diff
@@ -0,0 +1,176 @@
...
+import com.radixdlt.utils.UInt256;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+import static java.util.Objects.requireNonNull;
+
+public final class Transaction2DTO {
```
diegomrsantos, line 82: Same here as suggested earlier.

radixdlt-java/radixdlt-java/src/main/java/com/radixdlt/client/lib/dto/TransactionHistory2.java
```diff
@@ -0,0 +1,137 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.OptionalLong;
+
+import static java.util.Objects.requireNonNull;
+
+public final class TransactionHistory2 {
```
diegomrsantos, line 75: Same here as suggested earlier.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/accounts/BerkeleyAccountTxHistoryStore.java
talekhinezh: I think it's not applicable in this case since this iterator is strongly tied to the IO resource which needs to be closed after use. Trying to separate the functional part and the IO part before knowing the correct abstractions would be overdesign imo.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/module/ArchiveEndpointModule.java
talekhinezh: Yes, this will be resolved in a future PR.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/accounts/BerkeleyAccountTxHistoryStore.java
diegomrsantos: It is more about having small methods that tell what they are supposed to do very quickly from a high level perspective. Breaking it in smaller blocks that can be extracted to methods that describe what they are doing with their names.  It should be a private method. I think it would be even better if java supported nested methods.

```
public Stream<Pair<AID, Long>> getTxnIdsAssociatedWithAccount(REAddr addr, Long offset) {
		var cursor = accountTxHistory.openCursor(null, null);
		var iterator getIterator... (some descriiptive name)
		return Streams.stream(iterator)
			.onClose(cursor::close);
	}
```

siy (approved):
LGTM

PR #496, opened 2021-09-15
Feature Return Token RRI

siy:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/data/ApiErrors.java
```diff
@@ -102,6 +103,6 @@ public String message() {
 
 	@Override
 	public int code() {
-		return code;
+		return -2000 - ordinal();
```
diegomrsantos, line 106: I'd extract -`2000` into a variable with a name explaining what it means

michalzajda (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/data/PreparedTransaction.java
```diff
@@ -77,19 +81,22 @@
...
 		this.fee = fee;
+		this.notifications = notifications;
 	}
 
-	public static PreparedTransaction create(byte[] blob, byte[] hashToSign, UInt256 fee) {
+	public static PreparedTransaction create(byte[] blob, byte[] hashToSign, UInt256 fee, List<JSONObject> notifications) {
 		requireNonNull(blob);
 		requireNonNull(hashToSign);
 		requireNonNull(fee);
+		requireNonNull(notifications);
```
diegomrsantos, line 97: Can we add a message as a second parameter like "blob must not be null"?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/data/ApiErrors.java
siy: This is a preliminary step for reorganizing all errors - my next PR.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/data/PreparedTransaction.java
siy: This method is intended for use by Jackson, not by programmer. And in this scenario this check is even unnecessary as Jackson provides much more detailed error description when it happens.

diegomrsantos: If it is not necessary then I'd remove them.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/data/action/ResourceAction.java
```diff
@@ -0,0 +1,73 @@
...
+ * or non-commercial purpose and for any reproduction or redistribution by You of the
+ * Work. You assume all risks associated with Your use of the Work and the exercise of
+ * permissions under this License.
+ */
+
+package com.radixdlt.api.data.action;
+
+import com.radixdlt.identifiers.REAddr;
+
+public interface ResourceAction {
```
diegomrsantos, line 69: Could we make the name of it more descriptive or maybe add some docs? It not clear to me when classes should implement this interface.

diegomrsantos: Is it like `ActionableResource`? Meaning that any actionable resource should have a symbol and an address?

siy: Done

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/data/PreparedTransaction.java
siy: It's better to be safe than sorry. Someone might try to create instance manually and then check will be useful.

radixdlt-java/radixdlt-java/src/test/java/com/radixdlt/client/lib/api/sync/SyncRadixApiCreationTest.java
```diff
@@ -220,7 +220,16 @@ public void testCreateFixedSupplyToken() {
 			.onFailure(failure -> fail(failure.toString()))
 			.onSuccess(client -> client.transaction().build(request)
```
diegomrsantos, line 221: The formatting of this whole method is hard to understand imho.

diegomrsantos: I would do something like that 

```
		RadixApi.connect(BASE_URL)
			.map(RadixApi::withTrace)
			.onFailure(failure -> fail(failure.toString()))
			.onSuccess(client ->
				client.transaction().build(request)
				.onFailure(failure -> fail(failure.toString()))
				.onSuccess(builtTransaction -> {
					assertEquals(amount(100116600).micros(), builtTransaction.getFee());

					var notifications = builtTransaction.getTransaction().getNotifications();
					assertEquals(1, notifications.size());
					assertEquals("TokenCreate", notifications.get(0).getType());
					assertEquals("fix", notifications.get(0).getSymbol().orElse(""));
					assertEquals("fix_dr1qd5ah4xye2svl2smkk06st3q9lkqnkc6wgy0zxrrkmyszlawrz",
								 notifications.get(0).getRri().orElse(""));
				})
				.map(builtTransaction -> builtTransaction.toFinalized(KEY_PAIR1))
				.onSuccess(finalizedTransaction ->
					client.transaction().finalize(finalizedTransaction, true)
					.onFailure(failure -> fail(failure.toString()))
				)
			);
```
are we using a code formatter tool?

siy: Reformatted the whole class. Hopefully this will make it more readable.

siy: This is a matter of taste. Actually, initial formatting is more convenient and readable for me. Proposed one keeps parameter name on a separate line. This is a) inconsistent with a single line lambda formatting; b) requires to visually "lookup" parameter name, which is distracting.

diegomrsantos: Agree, not that good either. I was trying to make it easier to know when a method starts and finishes.

diegomrsantos (approved):

PR #498, opened 2021-09-17
New Error Handling Scheme

siy:


michalzajda (approved):

radixdlt-java-common/src/main/java/com/radixdlt/errors/RadixErrors.java
```diff
@@ -62,38 +63,97 @@
...
-	INVALID_UINT_VALUE("Invalid UInt256/UInt384 value {0}"),
+	INVALID_RESOURCE_ADDRESS("Invalid resource address {0}"),
+	INVALID_SIGNATURE_DER("Invalid signature DER {0}"),
+	INVALID_TX_ID("Invalid TX ID {0}"),
 	INVALID_VALIDATOR_ADDRESS("Invalid validator address {0}"),
-	OPERATION_INTERRUPTED("Operation interrupted with InterruptedException. Details: {0} {1}"),
+	INVALID_VALUE_OUT_OF_RANGE("Parameter {0} must be between {1} and {2}"),
+	INVALID_VALUE_TYPE("Invalid value type {0}"),
+
+	IO_ERROR("I/O Error: {0} {1}"),
```
talekhinezh, line 122: I don't think we should be defining errors which aren't related to the api here. Any internal server/io errors should collapse to some "Internal Server" error.

radixdlt-java-common/src/main/java/com/radixdlt/errors/RadixErrors.java
```diff
@@ -62,38 +63,97 @@
...
 
-package com.radixdlt.identifiers;
+package com.radixdlt.errors;
 
 import com.radixdlt.utils.functional.Failure;
 
-public enum CommonErrors implements Failure {
-	UNKNOWN_ERROR("Unknown error of type {0} with message {1}"),
+/**
+ * Full list of internal error codes.
```
talekhinezh, line 71: I don't think this should be a full list of internal codes, only codes which will be passed to api layer. Internal errors should be managed via different exception classes.

radixdlt-java-common/src/main/java/com/radixdlt/errors/RadixErrors.java
```diff
@@ -62,38 +63,97 @@
...
+	ERROR_PROCEDURE("{0}"),
+	ERROR_RESERVED_SYMBOL("{0}"),
+	ERROR_RESERVE_NOT_EMPTY("{0}"),
+	ERROR_RESOURCE_ALLOCATION_AND_DESTRUCTION("{0}"),
+	ERROR_SIGNED_SYSTEM("{0}"),
+	ERROR_SUBSTATE_NOT_FOUND("{0}"),
+	ERROR_VIRTUAL_PARENT_STATE_DOES_NOT_EXIST("{0}"),
+	ERROR_VIRTUAL_SUBSTATE_ALREADY_DOWN("{0}"),
+
+	INTERRUPTED_OPERATION("Operation interrupted with InterruptedException. Details: {0} {1}"),
```
talekhinezh, line 107: Also should not be pushing this error up to the API.

talekhinezh (changes requested):

radixdlt-java-common/src/main/java/com/radixdlt/errors/RadixErrors.java
siy: I believe that we should not use exceptions at all, as they're making code very hard to follow and reason about. But I agree that we should have codes separated for internal errors and errors which may appear at API level.

radixdlt-java-common/src/main/java/com/radixdlt/errors/RadixErrors.java
```diff
@@ -104,6 +164,6 @@ public String message() {
 
 	@Override
 	public int code() {
-		return -1000 - ordinal();
+		return Category.GENERAL.forId(ordinal());
```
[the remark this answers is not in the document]
siy: The idea is to eliminate the need to maintain error codes at all. As long as all new errors will be added at the end, this should not cause any issues.

radixdlt-java-common/src/main/java/com/radixdlt/errors/RadixErrors.java
siy: Moved to dedicated enum

radixdlt-java-common/src/main/java/com/radixdlt/errors/RadixErrors.java
siy: Moved to dedicated enum.

radixdlt-java-common/src/main/java/com/radixdlt/errors/RadixErrors.java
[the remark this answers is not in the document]
siy: Added error codes

PR #502, opened 2021-09-21
Feature/refactor validator api

talekhinezh:
* Refactor backend of validator api so it is no longer bottlenecked by database scans


michalzajda (approved):

siy (approved):
LGTM

PR #508, opened 2021-09-28
Switch to Java 17

siy:


rvelaz (approved):

LukasGasior1 (approved):

.github/workflows/add-artifacts-to-release.yml
```diff
@@ -24,10 +24,11 @@ jobs:
         with:
           ref: ${{ steps.set_environment.outputs.git_ref }}
           fetch-depth: 0
-      - name: Set up JDK 1.11
-        uses: actions/setup-java@v1
+      - name: Set up JDK 17
+        uses: actions/setup-java@v2
         with:
-          java-version: 1.11
+          distribution: 'zulu'
```
diegomrsantos, line 30: Should here and our docker files use the same distribution?

radixdlt-core/radixdlt/src/test/java/com/radixdlt/sync/SometimesByzantineCommittedReader.java
```diff
@@ -177,6 +179,7 @@ public VerifiedTxnsAndProof build() {
 		}
 	}
 
+	@Ignore
```
diegomrsantos, line 182: Could you please clarify why this is necessary?

.github/workflows/add-artifacts-to-release.yml
siy: Not necessarily. In fact, it might be even useful to check different distributions as this might help catch some incompatibilities which we would like to avoid.

radixdlt-core/radixdlt/src/test/java/com/radixdlt/sync/SometimesByzantineCommittedReader.java
siy: For some reason, a build on CI pipeline included this class into the list of tests and since the class is not test and not intended for such a use, the test failed and failed the whole build. Addition of this annotation solved the issue.

PR #516, opened 2021-10-18
Fix occasional duplicate sync request send

LukasGasior1:
# Pull Request Template

## Description

Fixes occasional sending of duplicate sync request as a result of obsolete ledger update timeout events.

### DIFF
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/sync/LocalSyncService.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/sync/LocalSyncService.java
@@ -242,7 +242,7 @@ public final class LocalSyncService {
 			))
 			.put(handler(
 				SyncingState.class, SyncLedgerUpdateTimeout.class,
-				state -> unused -> this.processSync(state)
+				state -> event -> this.processSyncLedgerUpdateTimeout(state, event)
 			))
 			.build();
 	}
@@ -422,7 +422,7 @@ public final class LocalSyncService {
 			);
 		} else {
 			this.syncLedgerUpdateTimeoutDispatcher.dispatch(
-				SyncLedgerUpdateTimeout.create(),
+				SyncLedgerUpdateTimeout.create(currentState.getCurrentHeader().getStateVersion()),
 				1000L
 			);
 			this.verifiedSender.sendVerifiedSyncResponse(syncResponse);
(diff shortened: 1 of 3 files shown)


radixdlt-core/radixdlt/src/main/java/com/radixdlt/sync/LocalSyncService.java
```diff
@@ -475,6 +475,14 @@ private SyncState processSyncRequestTimeout(SyncingState currentState, SyncReque
 		);
 	}
 
+	private SyncState processSyncLedgerUpdateTimeout(SyncingState currentState, SyncLedgerUpdateTimeout event) {
+		if (event.stateVersion() != currentState.getCurrentHeader().getStateVersion()) {
+			return currentState; // obsolete timeout event; ignore
+		} else {
+			return this.processSync(currentState);
```
dhedey, line 482: This looks like that this method is called when a ledger update isn't committed to the ledger within 1 second - and we need to recover from some error - in this case we call processSync to trigger sending of another sync request? Does this fit with your understanding? I'm wondering if we can add a quick comment explaining the intention?

dhedey: Possibly a bit of a weird request - but tbh, I think this class is a little hard to understand at present, and could do with a few more clearer names and comments. (Particular issues with some generic names "processSync", and overload of terms like "Sender"!), but small improvements here and there will help in time :)

dhedey: In fact, maybe just renaming the `processSync` method to `sendOffSyncRequestIfNeeded` and renaming `this.verifiedSender.sendVerifiedSyncResponse` to `this.verifiedSyncResponseHandler.handleVerifiedSyncResponse` and fixing up the class JavaDoc a little would make the whole class a lot clearer to me and we can avoid additional comments :D

LukasGasior1: I've renamed the verified/invalid sender in the newer PR (commit: https://github.com/radixdlt/radixdlt/pull/517/commits/b86d85bf2b1e50ae1b0c06861c784385f38f404e ) - good point, tbh I don't know why they were named like that ;)
I'm not so sure about processSync though, we call it in quite a few places and it does more than just dispatching the request

PR #517, opened 2021-10-18
Use round robin peers for sync requests

LukasGasior1:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/sync/SyncState.java
```diff
@@ -275,65 +275,66 @@ public int hashCode() {
 
 	final class SyncingState implements SyncState {
 		private final LedgerProof currentHeader;
-		private final ImmutableList<BFTNode> candidatePeers;
+		private final ImmutableList<BFTNode> candidatePeersQueue;
```
diegomrsantos, line 278: Do we need o rename this variable? I believe it is a good practice to avoid adding types to variable names.

PR #523, opened 2021-10-19
/health endpoint model update

Theoklitos:
The JSON body structure for `/health` changed and the tests need to be updated as well


siy (approved):
LGTM

PR #537, opened 2021-11-02
Get peer's remoteAddress in channelActive

LukasGasior1:
`socketChannel.remoteAddress()` might not always be available in the channel initializer. If it wasn't, we'll now try to get it in the channelActive callback.
This PR also removes unnecessary `throws` clause.

### DIFF
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/PeerChannel.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/PeerChannel.java
@@ -67,7 +67,6 @@ package com.radixdlt.network.p2p.transport;
 import com.google.common.util.concurrent.RateLimiter;
 import com.radixdlt.counters.SystemCounters;
 import com.radixdlt.crypto.ECKeyOps;
-import com.radixdlt.crypto.exception.PublicKeyException;
 import com.radixdlt.environment.EventDispatcher;
 import com.radixdlt.network.messaging.InboundMessage;
 import com.radixdlt.network.p2p.NodeId;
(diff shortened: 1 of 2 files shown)


siy (approved):
LGTM

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/PeerChannel.java
```diff
@@ -333,11 +334,11 @@ public boolean isOutbound() {
 	}
 
 	public String getHost() {
-		return this.host;
+		return remoteAddress.map(InetSocketAddress::getHostString).orElse("?");
 	}
 
 	public int getPort() {
-		return this.port;
+		return remoteAddress.map(InetSocketAddress::getPort).orElse(0);
```
diegomrsantos, line 341: What's the rationale for these default values?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/PeerChannel.java
```diff
@@ -333,11 +334,11 @@ public boolean isOutbound() {
 	}
 
 	public String getHost() {
-		return this.host;
+		return remoteAddress.map(InetSocketAddress::getHostString).orElse("?");
```
diegomrsantos, line 337: It makes using instance fields directly unsafe and it's not obvious.

diegomrsantos (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/PeerChannel.java
LukasGasior1: just not to have to deal with a null/optional everywhere else

PR #538, opened 2021-11-02
Fix deserialization of network data

siy:


radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/distributed/simulation/SimulationTest.java
```diff
@@ -298,13 +298,13 @@ public Builder numNodes(int numNodes, int numInitialValidators, Iterable<UInt256
 				.collect(ImmutableMap.toImmutableMap(ECKeyPair::getPublicKey, k -> stakesIterator.next()));
 
 			var initialVset = BFTValidatorSet.from(initialStakesMap.entrySet().stream()
-				.map(e -> BFTValidator.from(BFTNode.create(e.getKey()), e.getValue())));
+				.map(e -> BFTValidator.create(BFTNode.create(e.getKey()), e.getValue())));
```
diegomrsantos, line 301: I would rather use the naming convention that is used in JDK and recommended in the Effective Java book: https://www.fatalerrors.org/a/effective-java-chapter-1-preface-chapter-2-creating-and-destroying-objects.html

radixdlt-core/radixdlt/src/jmh/java/org/radix/benchmark/CodecBenchmark.java
```diff
@@ -66,17 +66,15 @@
 
 
 import com.radixdlt.DefaultSerialization;
+import com.radixdlt.serialization.ClassScanningSerializerIds;
+import com.radixdlt.serialization.DeserializeException;
+import com.radixdlt.serialization.DsonOutput.Output;
+import com.radixdlt.serialization.Serialization;
```
diegomrsantos, line 72: This seems only an import reordering and same happened to me when I used the _Optimize imports_ option. Do you know how we can fix it?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
```diff
@@ -307,7 +307,9 @@ private void processLedgerUpdate(LedgerUpdate ledgerUpdate) {
 	private void processEpochChange(EpochChange epochChange) {
 		// Sanity check
 		if (epochChange.getEpoch() != this.currentEpoch() + 1) {
-			throw new IllegalStateException("Bad Epoch change: " + epochChange + " current epoch: " + this.currentEpoch);
+			//TODO: looks like byzantine message, might be used for slashing?
```
diegomrsantos, line 310: Why was this exception removed?

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/distributed/simulation/SimulationTest.java
siy: The only alternative which has advantage is the factory method named after the class itself, like `bftValidator()` or `option()`. And the advantage, not big - factory method can be statically imported without risk of clashing with other similar imports. All other variants don't even have this advantage and basically equal to me.

radixdlt-core/radixdlt/src/jmh/java/org/radix/benchmark/CodecBenchmark.java
siy: Should we? Some time ago, both Eclipse and Idea were used for code formatting, hence the difference in the import ordering. It is possible to make Eclipse organize imports like Idea does, but not vice versa. Since we're using almost exclusively Idea now, over the time all files will be formatted identically. This file is ancient and was not touched for long time.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/DispatcherModule.java
```diff
@@ -419,26 +419,26 @@ private void configureSync() {
...
-				processors.forEach(e -> e.process(commit));
-				dispatcher.dispatch(commit);
+				processCommitSynchronously(commit, processors, systemCounters);
+				dispatcher.dispatch(commit);	//DISPATCH: processing looks safe
 			};
 		}
 	}
 
+	private void processCommitSynchronously(
+			BFTCommittedUpdate commit, @ProcessOnDispatch Set<EventProcessor<BFTCommittedUpdate>> processors, SystemCounters systemCounters
```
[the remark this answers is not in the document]
siy: fixed

radixdlt-core/radixdlt/src/main/java/com/radixdlt/DispatcherModule.java
```diff
@@ -419,26 +419,26 @@ private void configureSync() {
...
 			var dispatcher = environment.getDispatcher(BFTCommittedUpdate.class);
 			return commit -> {
-				long stateVersion = commit.getVertexStoreState().getRootHeader().getStateVersion();
-				systemCounters.set(CounterType.BFT_STATE_VERSION, stateVersion);
-				systemCounters.add(CounterType.BFT_PROCESSED, commit.getCommitted().size());
-				systemCounters.set(CounterType.BFT_VERTEX_STORE_SIZE, commit.getVertexStoreSize());
-				processors.forEach(e -> e.process(commit));
-				dispatcher.dispatch(commit);
+				processCommitSynchronously(commit, processors, systemCounters);
+				dispatcher.dispatch(commit);	//DISPATCH: processing looks safe
```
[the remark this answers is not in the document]
siy: removed

radixdlt-core/radixdlt/src/main/java/com/radixdlt/application/NodeApplication.java
```diff
@@ -109,9 +109,10 @@ private void processRequest(NodeApplicationRequest request) {
 			var txBuilder = radixEngine.construct(request.getRequest().feePayer(REAddr.ofPubKeyAccount(self)));
 			var txn = txBuilder.signAndBuild(hashSigner::sign);
 			var mempoolAdd = request.completableFuture()
-				.map(f -> MempoolAdd.create(txn, f))
+				.map(future -> MempoolAdd.create(txn, future))
 				.orElseGet(() -> MempoolAdd.create(txn));
-			this.mempoolAddEventDispatcher.dispatch(mempoolAdd);
+
+			this.mempoolAddEventDispatcher.dispatch(mempoolAdd);	//DISPATCH: event seems safe
```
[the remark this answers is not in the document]
siy: removed

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/BFTHeader.java
```diff
@@ -116,14 +113,20 @@ public BFTHeader(
...
-			HashUtils.zero256(),
-			ledgerHeader
+	@JsonCreator
+	public static BFTHeader create(
+		@JsonProperty(value = "view", required = true) long number,
+		@JsonProperty(value = "vertex_id", required = true) HashCode vertexId,
+		@JsonProperty(value = "ledger_header", required = true) LedgerHeader ledgerHeader
+	) {
+		return new BFTHeader(View.of(number), requireNonNull(vertexId), requireNonNull(ledgerHeader)
 		);
```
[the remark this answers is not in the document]
siy: Fixed

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
```diff
@@ -350,7 +352,8 @@ private void processConsensusEventInternal(ConsensusEvent consensusEvent) {
 		} else if (consensusEvent instanceof Vote) {
 			bftEventProcessor.processVote((Vote) consensusEvent);
 		} else {
-			throw new IllegalStateException("Unknown consensus event: " + consensusEvent);
+			//TODO: DISPATCH: add necessary branch once there will be more ConsensusEvent implementations
```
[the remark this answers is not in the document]
siy: done

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
```diff
@@ -307,7 +307,9 @@ private void processLedgerUpdate(LedgerUpdate ledgerUpdate) {
 	private void processEpochChange(EpochChange epochChange) {
 		// Sanity check
 		if (epochChange.getEpoch() != this.currentEpoch() + 1) {
-			throw new IllegalStateException("Bad Epoch change: " + epochChange + " current epoch: " + this.currentEpoch);
+			//TODO: looks like byzantine message, might be used for slashing?
+			log.warn("Bad Epoch change: {} current epoch: {}", epochChange, this.currentEpoch);
+			return;
```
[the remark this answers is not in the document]
siy: We received this message from network and this can be an attempt to crash the node. Why should we allow crashing, then?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
```diff
@@ -435,6 +438,7 @@ public void processBFTUpdate(BFTInsertUpdate update) {
 		};
 	}
 
+	//TODO: Why not just return syncTimeoutProcessor?
```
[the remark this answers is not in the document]
siy: added explanatory comment instead

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -249,7 +249,9 @@ public BFTSync(
 						this.vertexStore.highQC().highestCommittedQC(),
 						Optional.of(((FormedTC) viewQuorumReached.votingResult()).getTC()));
 			} else {
-				throw new IllegalArgumentException("Unknown voting result: " + viewQuorumReached.votingResult());
+				//TODO: DISPATCH: add necessary branch once there will be more ViewVotingResult implementations
```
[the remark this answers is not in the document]
siy: fixed

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -481,19 +484,19 @@ private void processVerticesResponseForCommittedSync(SyncState syncState, BFTNod
 		if (syncState.committedProof.getStateVersion() <= this.currentLedgerHeader.getStateVersion()) {
 			rebuildAndSyncQC(syncState);
 		} else {
-			ImmutableList<BFTNode> signers = syncState.committedProof.getSignersWithout(self);
+			var signers = syncState.committedProof.getSignersWithout(self);
```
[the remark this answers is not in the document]
siy: I think that final for local variables is an unnecessary noise.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/middleware2/network/ConsensusEventMessage.java
```diff
@@ -90,45 +90,44 @@
 	@DsonOutput(Output.ALL)
 	private final Vote vote;
 
-	ConsensusEventMessage() {
-		// Serializer only
-		this.proposal = null;
-		this.vote = null;
-	}
+	public ConsensusEventMessage(
+		@JsonProperty(value = "timestamp", required = true) long timestamp,
```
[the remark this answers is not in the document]
siy: removed

radixdlt-core/radixdlt/src/main/java/com/radixdlt/middleware2/network/ConsensusEventMessage.java
```diff
@@ -90,45 +90,44 @@
...
+		@JsonProperty("proposal") Proposal proposal,
+		@JsonProperty("vote") Vote vote
+	) {
+		super(timestamp);
+
+		if (proposal == null && vote == null) {
+			throw new IllegalStateException("No vote nor proposal are provided for ConsensusEventMessage");
```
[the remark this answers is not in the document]
siy: good point. fixed

radixdlt-core/radixdlt/src/main/java/com/radixdlt/middleware2/network/LedgerStatusUpdateMessage.java
```diff
@@ -78,18 +80,21 @@
...
 	@JsonProperty("header")
 	@DsonOutput(Output.ALL)
 	private final LedgerProof header;
 
-	LedgerStatusUpdateMessage() {
-		// Serializer only
-		this.header = null;
+	@JsonCreator
+	public LedgerStatusUpdateMessage(
+			@JsonProperty(value = "timestamp", required = true) long timestamp,
```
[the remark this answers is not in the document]
siy: fixed

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/discovery/DiscoverPeers.java
```diff
@@ -68,12 +68,13 @@
  * A message acting as as trigger for PeerDiscovery to start a discovery round.
  */
 public final class DiscoverPeers {
+	private static final DiscoverPeers INSTANCE = new DiscoverPeers();
```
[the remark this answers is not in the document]
siy: This eliminates unnecessary allocation of identical objects, which can't be distinguished anyway.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
siy: Because it will crash the node. Throwing an exception means that we trust the event received from network, which is not the case.

radixdlt-core/radixdlt/src/main/java/org/radix/network/messages/PeerPingMessage.java
```diff
@@ -64,13 +64,24 @@
...
 import org.radix.network.messaging.Message;
+import org.radix.time.Time;
 
 import java.util.Objects;
 
 @SerializerId2("p2p.liveness.ping")
 public final class PeerPingMessage extends Message {
+	@JsonCreator
+	public PeerPingMessage(@JsonProperty(value = "timestamp", required = true) long timestamp) {
+		super(timestamp);
```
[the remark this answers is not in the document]
siy: Good point. Need to look into it deeper.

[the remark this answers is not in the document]
siy: fixed

radixdlt-core/radixdlt/src/jmh/java/org/radix/benchmark/CodecBenchmark.java
diegomrsantos: I can cause many modifications in a file mostly related to import reordering if different people use different conventions.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
diegomrsantos: But it seems that's what we want to do in this case. @LukasGasior1, @talekhinezh?

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/distributed/simulation/SimulationTest.java
diegomrsantos: I still believe it is better to use the conventions of the JDK. We use `List.of()` not `List.create()`.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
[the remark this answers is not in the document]
diegomrsantos: Oh now I see what you mean. It seems two things can happen: either we received an invalid message from an honest node cause of an inconsistency in the consensus or an invalid message from a malicious node. How can we differentiate between them?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
diegomrsantos: It seems two things can happen: either we received an invalid message from an honest node cause of an inconsistency in the consensus or an invalid message from a malicious node. How can we differentiate between them?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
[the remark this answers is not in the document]
LukasGasior1: `LedgerUpdate` is a local message that's generated in StateComputer. It should never be invalid, regardless of what other message triggered the processing (sync response / consensus).

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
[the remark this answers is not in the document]
diegomrsantos: Maybe we could use polymorphism here and add a method `getHighQC(VertexStore vs)` to `ViewVotingResult`?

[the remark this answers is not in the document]
diegomrsantos:
```
public EventProcessor<ViewQuorumReached> viewQuorumReachedEventProcessor() {
		return viewQuorumReached -> {
			this.runOnThreads.add(Thread.currentThread().getName());

			final HighQC highQC = viewQuorumReached.votingResult().getHighQC(this.vertexStore);
			syncToQC(highQC, viewQuorumReached.lastAuthor());
		};
	}
```

```
public interface ViewVotingResult {

    static FormedQC qc(QuorumCertificate qc) {
        return new FormedQC(qc);
    }

    static FormedTC tc(TimeoutCertificate tc) {
        return new FormedTC(tc);
    }

    View getView();

    HighQC getHighQC(VertexStore vs);


    /**
     * Signifies that the view has been completed with a formed quorum certificate.
     */
    final class FormedQC implements ViewVotingResult {

        private final QuorumCertificate qc;

        public FormedQC(QuorumCertificate qc) {
            this.qc = qc;
        }

        public QuorumCertificate getQC() {
            return this.qc;
        }

        @Override
        public View getView() {
            return this.qc.getView();
        }

        @Override
        public HighQC getHighQC(VertexStore vs) {
            return HighQC.from(
                qc,
                vs.highQC().highestCommittedQC(),
                vs.getHighestTimeoutCertificate()
            );
        }

        @Override
        public boolean equals(Object o) {
            if (this == o) {
                return true;
            }
            if (o == null || getClass() != o.getClass()) {
                return false;
            }
            FormedQC formedQC = (FormedQC) o;
            return Objects.equals(qc, formedQC.qc);
        }

        @Override
        public int hashCode() {
            return Objects.hash(qc);
        }

        @Override
        public String toString() {
            return String.format("FormedQC{qc=%s}", qc);
        }
    }

    /**
     * Signifies that the view has been completed with a timeout certificate.
     */
    final class FormedTC implements ViewVotingResult {

        private final TimeoutCertificate tc;

        public FormedTC(TimeoutCertificate tc) {
            this.tc = tc;
        }

        public TimeoutCertificate getTC() {
            return this.tc;
        }

        @Override
        public View getView() {
            return this.tc.getView();
        }

        @Override
        public HighQC getHighQC(VertexStore vs) {
            return HighQC.from(
                vs.highQC().highestQC(),
                vs.highQC().highestCommittedQC(),
                Optional.of(tc)
            );
        }

        @Override
        public boolean equals(Object o) {
            if (this == o) {
                return true;
            }
            if (o == null || getClass() != o.getClass()) {
                return false;
            }
            FormedTC formedTC = (FormedTC) o;
            return Objects.equals(tc, formedTC.tc);
        }

        @Override
        public int hashCode() {
            return Objects.hash(tc);
        }

        @Override
        public String toString() {
            return String.format("FormedTC{tc=%s}", tc);
        }
    }
}


```

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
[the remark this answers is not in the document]
diegomrsantos: In this case I believe we should keep the exception.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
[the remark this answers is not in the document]
diegomrsantos: Seems we could do something similar here and merge `bftEventProcessor.processVote` and `bftEventProcessor.processProposal `to `bftEventProcess(ConsensusEvent)`

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/BFTHeader.java
```diff
@@ -116,12 +113,17 @@ public BFTHeader(
 		this.ledgerHeader = ledgerHeader;
 	}
 
+	@JsonCreator
+	public static BFTHeader create(
+		@JsonProperty(value = "view", required = true) long number,
+		@JsonProperty(value = "vertex_id", required = true) HashCode vertexId,
+		@JsonProperty(value = "ledger_header", required = true) LedgerHeader ledgerHeader
+	) {
+		return new BFTHeader(View.of(number), requireNonNull(vertexId), requireNonNull(ledgerHeader));
```
diegomrsantos, line 122: I would move validation to constructor as it is public thus still possible to create an invalid Instance.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/HighQC.java
```diff
@@ -159,7 +161,7 @@ public static HighQC from(
 		QuorumCertificate highestCommittedQC,
 		Optional<TimeoutCertificate> highestTC
 	) {
-		return new HighQC(highestQC, Objects.requireNonNull(highestCommittedQC), highestTC.orElse(null));
+		return new HighQC(highestQC, requireNonNull(highestCommittedQC), highestTC.orElse(null));
```
diegomrsantos, line 164: Constructor accepts null `highestCommittedQC`, but here we require it to be non null.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/LedgerHeader.java
```diff
@@ -123,11 +125,11 @@
 	private LedgerHeader(
 		@JsonProperty("epoch") long epoch,
 		@JsonProperty("view") long view,
-		@JsonProperty("accumulator_state") AccumulatorState accumulatorState,
+		@JsonProperty(value = "accumulator_state", required = true) AccumulatorState accumulatorState,
 		@JsonProperty("timestamp") long timestamp,
 		@JsonProperty("next_validators") ImmutableSet<BFTValidator> nextValidators
 	) {
-		this(epoch, View.of(view), accumulatorState, timestamp, nextValidators);
+		this(epoch, View.of(view), requireNonNull(accumulatorState), timestamp, nextValidators);
```
diegomrsantos, line 132: I would move validation to constructor.

radixdlt-core/radixdlt/src/jmh/java/org/radix/benchmark/CodecBenchmark.java
siy: Yes. But we don't commit files without other changes. In this case, changes in imports is irrelevant.

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/distributed/simulation/SimulationTest.java
siy: Actual naming is mostly irrelevant as long as it is consistent. Changing naming in this PR will make it less consistent.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/TimestampedECDSASignatures.java
```diff
@@ -162,7 +164,7 @@ public TimestampedECDSASignatures() {
 	 * Returns a new instance containing {@code nodeToTimestampAndSignature}.
 	 *
 	 * @param nodeToTimestampAndSignature The map of {@link com.radixdlt.crypto.ECDSASignature}s and their corresponding
-	 * 	timestamps and {@link com.radixdlt.crypto.ECPublicKey}
+	 *                                    timestamps and {@link com.radixdlt.crypto.ECPublicKey}
 	 */
 	public TimestampedECDSASignatures(Map<BFTNode, TimestampedECDSASignature> nodeToTimestampAndSignature) {
```
diegomrsantos, line 169: It is still possible to create an invalid instance using the constructor.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/TimestampedECDSASignatures.java
```diff
@@ -111,7 +111,9 @@
 	private Map<BFTNode, TimestampedECDSASignature> nodeToTimestampedSignature;
 
 	@JsonCreator
-	public static TimestampedECDSASignatures from(@JsonProperty("signatures") Map<String, TimestampedECDSASignature> signatures) {
+	public static TimestampedECDSASignatures from(
+		@JsonProperty("signatures") Map<String, TimestampedECDSASignature> signatures
```
diegomrsantos, line 115: If we pass a null value this method fails cause of an implicit requireNonNull in the collector. Would we like to add an explicit check in our code?

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/distributed/simulation/SimulationTest.java
diegomrsantos: I believe naming is still relevant, java developers will be familiar with conventions used in the ecosystem.  Additionally changing few places to `create` won't make it consistent in the code as there are other places still using `from`.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/UnverifiedVertex.java
```diff
@@ -155,7 +163,7 @@ public static UnverifiedVertex create(
 		List<Txn> txns,
 		BFTNode proposer
 	) {
-		Objects.requireNonNull(qc);
+		requireNonNull(qc);
```
diegomrsantos, line 166: This is not necessary as this validation happens in the constructor.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/UnverifiedVertex.java
```diff
@@ -155,7 +163,7 @@ public static UnverifiedVertex create(
...
 
 		if (view.number() == 0) {
```
diegomrsantos, line 168: Should be moved to view constructor.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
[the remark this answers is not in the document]
LukasGasior1: I'm personally not a big fan of adding methods that depend on other services in the "dto" classes.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/discovery/DiscoverPeers.java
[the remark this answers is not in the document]
LukasGasior1: yeah, right, but then it requires additional work to add any fields in the future

radixdlt-core/radixdlt/src/main/java/com/radixdlt/ledger/DtoTxnsAndProof.java
```diff
@@ -105,8 +105,8 @@
 	@JsonCreator
 	public DtoTxnsAndProof(
 		@JsonProperty("txns") List<byte[]> txns,
-		@JsonProperty("head") DtoLedgerProof head,
-		@JsonProperty("tail") DtoLedgerProof tail
+		@JsonProperty(value = "head", required = true) DtoLedgerProof head,
+		@JsonProperty(value = "tail", required = true) DtoLedgerProof tail
 	) {
 		this.txns = txns == null ? ImmutableList.of() : txns;
```
diegomrsantos, line 111: We aren't checking if the list contains nulls.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/discovery/DiscoverPeers.java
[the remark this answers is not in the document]
siy: We always can return to regular instances, when this will be necessary.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/BFTHeader.java
siy: done

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/HighQC.java
siy: done

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/LedgerHeader.java
siy: done

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/TimestampedECDSASignatures.java
siy: fixed

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/TimestampedECDSASignatures.java
siy: If we pass `null` here, it will be properly handled. Or you mean that here we should check map internals?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/UnverifiedVertex.java
siy: done

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/UnverifiedVertex.java
siy: I don't think so. We don't allow it here, but genesis view is 0 and this is valid case.

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/distributed/simulation/SimulationTest.java
siy: Please add relevant ticket into backlog. Doing such a renaming in the context of current PR is not feasible since has nothing to do with the task and will affect a number of files which are not target of the fix.

diegomrsantos: I wasn't suggesting to do an extensive renaming in this task, but actually revert the renaming that has been done. 
Ticket has been created for renaming https://radixdlt.atlassian.net/browse/RPNV1-1527

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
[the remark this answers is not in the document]
diegomrsantos: I agree, but overuse of instanceOf also seems a code smell. It violates the open/closed principle and in a language that doesn't have sealed classes, there is no way to know which if/else or switch blocks should be update when new implementations are added. The only two solutions I know for this is the Visitor Pattern and Polymorphism.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/TimestampedECDSASignatures.java
diegomrsantos: I meant when we receive a `Map` with `TimestampedECDSASignature` value as null.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/UnverifiedVertex.java
diegomrsantos: makes sense

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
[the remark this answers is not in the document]
siy: Reworked using "inverse dispatch". Safe and hard to miss if new event will be added.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/middleware2/network/ConsensusEventMessage.java
```diff
@@ -90,45 +89,45 @@
...
-	}
+	public ConsensusEventMessage(
+		@JsonProperty("proposal") Proposal proposal,
+		@JsonProperty("vote") Vote vote
+	) {
+		if (proposal == null && vote == null) {
+			throw new IllegalStateException("No vote nor proposal are provided for ConsensusEventMessage");
+		}
+
+		if (proposal != null && vote != null) {
```
diegomrsantos, line 103: I believe here we should have `||` instead.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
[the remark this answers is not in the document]
LukasGasior1: or just wait for java 17 and seal the interface :)

radixdlt-core/radixdlt/src/main/java/com/radixdlt/middleware2/network/GetVerticesRequestMessage.java
```diff
@@ -85,13 +87,11 @@
...
-	}
-
-	GetVerticesRequestMessage(HashCode vertexId, int count) {
+	@JsonCreator
+	public GetVerticesRequestMessage(
+		@JsonProperty(value = "vertexId", required = true) HashCode vertexId,
+		@JsonProperty(value = "count", required = true) int count
+	) {
 		this.vertexId = Objects.requireNonNull(vertexId);
 		this.count = count;
```
diegomrsantos, line 96: We need to check count is valid here, that was one of the problems that @LukasGasior1 found. @LukasGasior1 should it be greater than zero or it can be zero too?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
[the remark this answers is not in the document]
diegomrsantos: Seems a better solution

radixdlt-core/radixdlt/src/main/java/com/radixdlt/middleware2/network/GetVerticesRequestMessage.java
diegomrsantos: Given that, I believe we need to add the same validation in `GetVerticesRequest`

radixdlt-core/radixdlt/src/main/java/com/radixdlt/middleware2/network/ConsensusEventMessage.java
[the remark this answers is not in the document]
diegomrsantos: Created a ticket to address that https://radixdlt.atlassian.net/browse/RPNV1-1528

radixdlt-core/radixdlt/src/main/java/com/radixdlt/DispatcherModule.java
```diff
@@ -419,26 +419,28 @@ private void configureSync() {
...
-				systemCounters.add(CounterType.BFT_PROCESSED, commit.getCommitted().size());
-				systemCounters.set(CounterType.BFT_VERTEX_STORE_SIZE, commit.getVertexStoreSize());
-				processors.forEach(e -> e.process(commit));
+				processCommitSynchronously(commit, processors, systemCounters);
 				dispatcher.dispatch(commit);
 			};
 		}
 	}
 
+	private void processCommitSynchronously(
```
talekhinezh, line 432: Helpful if `static` is added.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/middleware2/network/GetVerticesResponseMessage.java
```diff
@@ -84,17 +84,15 @@
...
-		// Serializer only
-		this.vertices = null;
-	}
-
-	GetVerticesResponseMessage(List<UnverifiedVertex> vertices) {
+	@JsonCreator
+	public GetVerticesResponseMessage(
+		@JsonProperty(value = "vertices", required = true) List<UnverifiedVertex> vertices
+	) {
 		this.vertices = Objects.requireNonNull(vertices);
```
diegomrsantos, line 91: We aren't checking for nulls inside the list.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/middleware2/network/GetVerticesResponseMessage.java
```diff
@@ -84,17 +84,15 @@
...
 
-	GetVerticesResponseMessage() {
-		// Serializer only
-		this.vertices = null;
-	}
-
-	GetVerticesResponseMessage(List<UnverifiedVertex> vertices) {
+	@JsonCreator
+	public GetVerticesResponseMessage(
+		@JsonProperty(value = "vertices", required = true) List<UnverifiedVertex> vertices
```
diegomrsantos, line 89: We aren't checking for nulls inside the list.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/application/NodeApplication.java
```diff
@@ -109,8 +109,9 @@ private void processRequest(NodeApplicationRequest request) {
 			var txBuilder = radixEngine.construct(request.getRequest().feePayer(REAddr.ofPubKeyAccount(self)));
 			var txn = txBuilder.signAndBuild(hashSigner::sign);
 			var mempoolAdd = request.completableFuture()
-				.map(f -> MempoolAdd.create(txn, f))
+				.map(future -> MempoolAdd.create(txn, future))
```
talekhinezh, line 112: Would prefer to not have arbitrary decoration changes in non-relevant classes. But fine for now.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/BFTHeader.java
```diff
@@ -98,30 +100,28 @@
...
-		this.vertexId = vertexId;
-		this.ledgerHeader = ledgerHeader;
+		this.view = requireNonNull(view);
+		this.vertexId = requireNonNull(vertexId);
+		this.ledgerHeader = requireNonNull(ledgerHeader);
+	}
+
+	@JsonCreator
+	public static BFTHeader create(
+		@JsonProperty("view") long number,
```
talekhinezh, line 116: This should be `required` as well?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/ConsensusEvent.java
```diff
@@ -97,4 +97,11 @@
 	 * @return {@link HighQC} associated with event
 	 */
 	HighQC highQC();
+
+	/**
+	 * Dispatch processing to corresponding method in the processor.
+	 *
+	 * @param processor	the processor to use
+	 */
+	void processBy(BFTEventProcessor processor);
```
talekhinezh, line 106: `ConsensusEvent` should not know about who's processing it or how's it being processed. This breaks abstractions. Please remove.

talekhinezh (changes requested):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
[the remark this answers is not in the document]
diegomrsantos: It is better, but a lot of if/else's ins't cool. Using new switch with pattern matching makes it even better, but it still violates the open/close principle. If we are fine with that, so not an issue.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/ConsensusEvent.java
diegomrsantos: But it still doesn't know, BFTEventProcessor is also an abstraction. This is some kind of simpler Visitor pattern to avoid overuse of `if/else` and `instanceOf`. With a sealed interface and pattern matching it becomes less critical though.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/BFTHeader.java
diegomrsantos: This a good question. I think it was assumed to be already required as it is a primitive type, but not sure it is a correct assumption as the default value could be used instead.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -369,7 +371,8 @@ private void processGetVerticesLocalTimeout(VertexRequestTimeout timeout) {
 		}
 
 		if (syncRequestState.authors.isEmpty()) {
-			throw new IllegalStateException("Request contains no authors except ourselves");
+			log.warn("Request contains no authors except ourselves");
```
talekhinezh, line 374: This is a valid `IllegalStateException` since `VertexRequestTimeout` is a local message.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/environment/Dispatchers.java
```diff
@@ -79,6 +79,7 @@
 /**
  * Helper class to set up environment with dispatched events
  */
+//TODO: get rid of field injection
```
talekhinezh, line 82: Why?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/middleware2/network/GetVerticesResponseMessage.java
talekhinezh: Depends on whether nulls can be injected by our serializer (aka Cbor).

radixdlt-core/radixdlt/src/main/java/com/radixdlt/middleware2/network/MessageCentralBFTNetwork.java
```diff
@@ -96,23 +95,21 @@ public MessageCentralBFTNetwork(MessageCentral messageCentral) {
...
-				return RemoteEvent.create(node, proposal);
-			});
+			.map(this::toRemoteEvent);
+	}
+
+	@SuppressWarnings("unchecked")
+	private <T extends ConsensusEvent> RemoteEvent<T> toRemoteEvent(MessageFromPeer<ConsensusEventMessage> m) {
+		final var node = BFTNode.create(m.getSource().getPublicKey());
+		final var msg = m.getMessage();
+		var vote = (T) msg.getConsensusMessage();
```
talekhinezh, line 111: change `vote` to `t`

radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/PersistenceModule.java
```diff
@@ -115,7 +115,7 @@ StoreConfig storeConfig() {
 		return update -> {
 			systemCounters.increment(CounterType.PERSISTENCE_VERTEX_STORE_SAVES);
 			persistentVertexStore.save(update.getVertexStoreState());
-		};
+		};	//TODO: DISPATCH: processing PersistentVertexStore may throw BerkeleyStoreException
```
talekhinezh, line 118: `BerkeleyStoreException` is a true `panic` exception so it shouldn't be handled.

radixdlt-core/radixdlt/src/test/java/org/radix/serialization/SerializeObject.java
```diff
@@ -112,29 +108,13 @@ protected SerializeObject(Class<T> cls, Supplier<T> factory) {
...
-	}
-
-	@Test
-	public void testNONEIsEmpty() {
-		String s2Json = getSerialization().toJson(factory.get(), Output.NONE);
-		assertEquals("{}", s2Json);
-	}
-
-	@Test
-	public void testRoundTripJsonSame() throws DeserializeException {
```
talekhinezh, line 131: Not sure why these were removed?

radixdlt-core/radixdlt/src/test/java/org/radix/serialization/Serializer2Test.java
```diff
@@ -119,21 +115,6 @@ public void roundTripJacksonDsonTest() throws DeserializeException {
 		assertEquals(testObject, newObject);
 	}
 
-	@Test
-	public void roundTripJacksonJsonTest() throws DeserializeException {
```
talekhinezh, line 123: Similarly why were these removed?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/middleware2/network/ConsensusEventMessage.java
```diff
@@ -90,45 +92,37 @@
...
-		if (this.vote != null) {
-			return this.vote;
-		}
+	public ConsensusEventMessage(Vote vote) {
+		this(null, vote);
+	}
 
-		return null;
+	public ConsensusEvent getConsensusMessage() {
+		return requireNonNull(coalesce(proposal, vote));
```
talekhinezh, line 120: I'd rather not remove the `if else` statements. It's wrong but we shouldn't hide the antipattern and resolve the issue in the correct way.

talekhinezh (changes requested):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/middleware2/network/GetVerticesResponseMessage.java
diegomrsantos: But we can't assume we will receive only message serialised but the system, can we?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/UnverifiedVertex.java
```diff
@@ -109,30 +111,36 @@
...
+		this.qc = requireNonNull(qc);
+		this.view = requireNonNull(view);
 
 		if (proposerTimedOut != null && proposerTimedOut && !txns.isEmpty()) {
 			throw new IllegalArgumentException("Txns must be empty if timeout");
 		}
 
+		if (txns != null && !txns.isEmpty()) {
+			for (var txn : txns) {
+				requireNonNull(txn);
```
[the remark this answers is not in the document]
diegomrsantos: But we can't assume we will receive only message serialised but the system, can we?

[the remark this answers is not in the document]
diegomrsantos: Ah I think you mean the deserialisation process won't create a List with null inside? But this isn't the only way of creating the object and allowing this could cause unexpected behaviour in the system.

radixdlt-core/radixdlt/src/main/java/org/radix/network/messaging/Message.java
```diff
@@ -96,7 +94,7 @@ protected Message() {
 		this(Time.currentTimestamp());
 	}
 
-	protected Message(long timestamp) {
+	private Message(long timestamp) {
```
diegomrsantos, line 97: Why is this private? Additionally, use of `Time.currentTimestamp()` will probably make this class untestable.

radixdlt-core/radixdlt/src/test/java/com/radixdlt/ConsensusModuleTest.java
```diff
@@ -172,7 +172,7 @@ public void setup() {
 		var genesis = UnverifiedVertex.createGenesis(LedgerHeader.genesis(accumulatorState, null, 0));
 		var hashedGenesis = new VerifiedVertex(genesis, HashUtils.zero256());
 		var qc = QuorumCertificate.ofGenesis(hashedGenesis, LedgerHeader.genesis(accumulatorState, null, 0));
-		var validatorSet = BFTValidatorSet.from(Stream.of(BFTValidator.from(BFTNode.random(), UInt256.ONE)));
+		var validatorSet = BFTValidatorSet.from(Stream.of(BFTValidator.create(BFTNode.random(), UInt256.ONE)));
```
diegomrsantos, line 175: Can we please revert that?

radixdlt-core/radixdlt/src/test/java/com/radixdlt/consensus/bft/BFTValidatorSetTest.java
```diff
@@ -64,25 +64,22 @@
...
-import java.util.stream.IntStream;
-
-import org.junit.Test;
-
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import com.radixdlt.crypto.ECKeyPair;
 import com.radixdlt.crypto.HashUtils;
+import com.radixdlt.utils.UInt256;
```
diegomrsantos, line 72: I'd revert all changes in this file as they are not necessary.

radixdlt-core/radixdlt/src/test/java/com/radixdlt/consensus/liveness/WeightedRotatingLeadersTest.java
```diff
@@ -172,7 +172,7 @@ public void when_validators_distributed_by_fibonacci__then_leaders_also_distribu
 
 		final int sumOfPower = fibonacci.get().sum();
 		this.validatorsInOrder = fibonacci.get()
-			.mapToObj(p -> BFTValidator.from(BFTNode.random(), UInt256.from(p)))
+			.mapToObj(p -> BFTValidator.create(BFTNode.random(), UInt256.from(p)))
```
diegomrsantos, line 175: Same here and all other places, I believe there's no reason to change that.

radixdlt-core/radixdlt/src/test/java/com/radixdlt/middleware2/network/ConsensusEventMessageTest.java
```diff
@@ -104,21 +103,6 @@ public void sensibleToStringVote() {
 		assertTrue(msg1.getConsensusMessage() instanceof Vote);
 	}
 
-	@Test
```
diegomrsantos, line 107: Why they were removed?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/middleware2/network/GetVerticesResponseMessage.java
diegomrsantos: Ah I think you mean the deserialisation process won't create a List with null inside? But this isn't the only way of creating the object and allowing this could cause unexpected behaviour in the system.

siy:
PR grew far beyond initial scope

PR #545, opened 2021-11-06
Update parsing.md

MRumpold:
## Description

Currently the 'owner' datatype is specified as u256, I think it should be address.

In the substate TOKEN_RESOURCE the reserved u8 is marked with 'type' at the beginning, in all others with 'reserved'.

In the description of the optional fields it looks to me like no data should follow for '0x00', but when I look at the transactions in the radix.ledger file, data follows with the 'some' length and value '00...'.

## Labels

docs


iamyulong (approved):
Thanks for providing the fix!

PR #549, opened 2021-11-10
[DO-122] Enable parallel run on integration tests

muzuke:


muzuke:
This doesn't seem to have any effect. The build times are the same with previous runs.

muzuke:
@Theoklitos @diegomrsantos Is there any other way to run the integrations tests in parallel?

muzuke:
@diegomrsantos Is there anything that can be done on the CI configuration to leverage more cores? Or should I close this PR?

diegomrsantos:
@muzuke I think @rvelaz mentioned one day that we will need to have our own runners for that.

rvelaz:
We can merge the change, but it won't have any effect until we use different runners

diegomrsantos:
Is this change necessary?

diegomrsantos:
Can we close this pr?

PR #550, opened 2021-11-11
Feature/missing network changes

LukasGasior1:
Brings back these 3 PRs:
* https://github.com/radixdlt/radixdlt/pull/455
* https://github.com/radixdlt/radixdlt/pull/473
* https://github.com/radixdlt/radixdlt/pull/474

### DIFF
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBook.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBook.java
@@ -235,4 +235,13 @@ public final class AddressBook {
 			}
 		}
 	}
+
+	public void clear() {
+		synchronized (lock) {
+			this.persistence.close();
+			this.persistence.reset();
+			this.persistence.open();
+			this.knownPeers.clear();
+		}
+	}
 }
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBook.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBook.java
@@ -74,10 +74,12 @@ import com.radixdlt.network.p2p.PeerEvent;
 import com.radixdlt.network.p2p.PeerEvent.PeerBanned;
 import com.radixdlt.network.p2p.RadixNodeUri;
 import com.radixdlt.network.p2p.addressbook.AddressBookEntry.PeerAddressEntry;
+import com.radixdlt.network.p2p.addressbook.AddressBookEntry.PeerAddressEntry.LatestConnectionStatus;
 
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
@@ -91,17 +93,51 @@ import static java.util.function.Predicate.not;
  * Manages known peers network addresses and their metadata.
  */
 public final class AddressBook {
-	private static final Comparator<AddressBookEntry.PeerAddressEntry> entryComparator = (a, b) -> {
-		final var aLastSuccess = a.getLastSuccessfulConnection().orElse(Instant.MIN);
-		final var bLastSuccess = b.getLastSuccessfulConnection().orElse(Instant.MIN);
-		return -aLastSuccess.compareTo(bLastSuccess);
-	};
+
+	/**
+	 * A stateful comparator for known peer addresses that uses both their latest connection status (persistent)
+	 * and a number of failed connection attempts (volatile). Failure counts are used to cycle the addresses for retries.
+	 * The entries are sorted in the following manner:
+	 * successful -> unknown (no connection attempted yet) -> failed (by num of failures)
+	 */
+	private static final class PeerAddressEntryComparator implements Comparator<PeerAddressEntry> {
+		private final Map<RadixNodeUri, Integer> failureCounts = new HashMap<>();
+
+		private int toIntValue(PeerAddressEntry peerAddressEntry) {
+			return peerAddressEntry.getLatestConnectionStatus().map(latestConnectionStatus -> {
+				if (latestConnectionStatus == LatestConnectionStatus.SUCCESS) {
+					return 1;
+				} else {
+					return -(1 + failureCounts.getOrDefault(peerAddressEntry.getUri(), 0));
+				}
+			}).orElse(0);
+		}
+
+		@Override
+		public int compare(PeerAddressEntry a, PeerAddressEntry b) {
+			return Integer.compare(toIntValue(b), toIntValue(a));
+		}
+
+		void incFailures(RadixNodeUri uri) {
+			synchronized (this.failureCounts) {
+				final var curr = this.failureCounts.getOrDefault(uri, 0);
+				this.failureCounts.put(uri, curr + 1);
+			}
+		}
+
+		void resetFailures(RadixNodeUri uri) {
+			synchronized (this.failureCounts) {
+				this.failureCounts.remove(uri);
+			}
+		}
+	}
 
 	private final RadixNodeUri self;
 	private final EventDispatcher<PeerEvent> peerEventDispatcher;
 	private final AddressBookPersistence persistence;
 	private final Object lock = new Object();
 	private final Map<NodeId, AddressBookEntry> knownPeers = new ConcurrentHashMap<>();
+	private final PeerAddressEntryComparator addressEntryComparator = new PeerAddressEntryComparator();
 
 	@Inject
 	public AddressBook(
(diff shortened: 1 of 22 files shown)


dhedey (approved):
Looks good to me

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBook.java
```diff
@@ -91,35 +93,95 @@
...
 
-	private final NodeId selfNodeId;
+	/**
+	 * A stateful comparator for known peer addresses that uses both their latest connection status (persistent)
+	 * and a number of failed connection attempts (volatile). Failure counts are used to cycle the addresses for retries.
+	 * The entries are sorted in the following manner:
+	 * successful -> unknown (no connection attempted yet) -> failed (by num of failures)
+	 */
+	private static final class PeerAddressEntryComparator implements Comparator<PeerAddressEntry> {
+		private final Map<RadixNodeUri, Integer> failureCounts = new HashMap<>();
```
diegomrsantos, line 104: What do you think about moving this to `PeerAddressEntry`? Not sure it is the Comparator responsibility.

diegomrsantos: @LukasGasior1 answer: I was considering adding it to the address book entry and persisting (failureCount), but I think it's not really required to be persisted and keeping it in memory saves a db call

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
```diff
@@ -121,8 +122,8 @@ public static AddressBookEntry createBanned(NodeId nodeId, Instant bannedUntil)
 		return new AddressBookEntry(nodeId, Optional.of(bannedUntil), ImmutableSet.of());
 	}
 
-	public static AddressBookEntry create(RadixNodeUri uri, Instant lastSuccessfulConnection) {
-		return create(uri, Optional.of(lastSuccessfulConnection));
+	public static AddressBookEntry createWithLatestConnectionStatus(RadixNodeUri uri, LatestConnectionStatus latestConnectionStatus) {
+		return create(uri, Optional.of(latestConnectionStatus));
```
diegomrsantos, line 126: Maybe Optional.ofNullable?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
```diff
@@ -133,11 +134,11 @@ public static AddressBookEntry createBlacklisted(RadixNodeUri uri, Instant black
 		);
 	}
 
-	public static AddressBookEntry create(RadixNodeUri uri, Optional<Instant> lastSuccessfulConnection) {
+	public static AddressBookEntry create(RadixNodeUri uri, Optional<LatestConnectionStatus> latestConnectionStatus) {
 		return new AddressBookEntry(
 			uri.getNodeId(),
 			Optional.empty(),
-			ImmutableSet.of(new AddressBookEntry.PeerAddressEntry(uri, lastSuccessfulConnection, Optional.empty()))
+			ImmutableSet.of(new AddressBookEntry.PeerAddressEntry(uri, latestConnectionStatus, Optional.empty()))
```
diegomrsantos, line 141: It is probably gonna fail with NullPointerException if latestConnectionStatus is null. Is that what we want?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
```diff
@@ -210,7 +211,7 @@ public AddressBookEntry withLastSuccessfulConnectionFor(RadixNodeUri uri, Instan
 				.build();
 			return new AddressBookEntry(nodeId, bannedUntil, newKnownAddresses);
 		} else {
-			final var newAddressEntry = new PeerAddressEntry(uri, Optional.of(lastSuccessfulConnectionFor), Optional.empty());
+			final var newAddressEntry = new PeerAddressEntry(uri, Optional.of(latestConnectionStatus), Optional.empty());
```
diegomrsantos, line 214: Maybe Optional.ofNullable?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
```diff
@@ -288,33 +293,33 @@ public int hashCode() {
...
 
 		@JsonCreator
 		private static PeerAddressEntry deserialize(
 			@JsonProperty("uri") RadixNodeUri uri,
-			@JsonProperty("lastSuccessfulConnection") Long rawLastSuccessfulConnection,
+			@JsonProperty("latestConnectionStatus") String rawLatestConnectionStatus,
 			@JsonProperty("blacklistedUntil") Long rawBlacklistedUntil
 		) {
-			final var lastSuccessfulConnection = Optional.ofNullable(rawLastSuccessfulConnection).map(Instant::ofEpochMilli);
+			final var latestConnectionStatus = Optional.ofNullable(rawLatestConnectionStatus).map(LatestConnectionStatus::valueOf);
```
diegomrsantos, line 306: what if `valueOf` fails?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
LukasGasior1: I think we should only use `ofNullable` when the null value is an expected valid case (like, for example, the `deserialize` method that has to deal with previously persisted entries). In general we shouldn't expose any nulls (or methods that accept nulls) outside the local scope.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
LukasGasior1: I think so, it should never be null

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
LukasGasior1: it'll fail with a runtime exception, do you think we should handle it differently?
this class is not deserialized from an

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
diegomrsantos: We can't prevent user from passing null. The way it is now a null pointer exception is gonna be thrown. Do we want to do that or fallback to Optional.empty()?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
diegomrsantos: I believe it is better to enforce that explicitly.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
diegomrsantos: Not sure we can assume the string will always be correct. if it is an optional field, we could log it as a warn and fallback to Optinal.empty() instead of throwing an exception. What do you think?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
```diff
@@ -325,10 +330,10 @@ public boolean blacklistExpired() {
 			return blacklistedUntil.isPresent() && !blacklisted();
 		}
 
-		@JsonProperty("lastSuccessfulConnection")
+		@JsonProperty("latestConnectionStatus")
 		@DsonOutput(DsonOutput.Output.ALL)
-		private Long getLastSuccessfulConnectionForSerializer() {
-			return lastSuccessfulConnection.map(Instant::toEpochMilli).orElse(null);
+		private String getLatestConnectionStatusForSerializer() {
```
diegomrsantos, line 335: Can't it handle Optional?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
LukasGasior1: I'd rather have an NPE in this case, we just shouldn't be using nulls

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
LukasGasior1: so that's a broader question, should we add `Objects.requireNonNull` everywhere where null is not expected?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
LukasGasior1: nope

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
diegomrsantos: Maybe it is a good idea. Should we adopt this approach for the whole codebase?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
LukasGasior1: I'm not sure, I think I'd rather fail with an error cause this would mean there's either a database corruption or a serious serialization bug

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
LukasGasior1: we already do that, other than a few local uses and some interfaces that are explicitly marked with `@Nullable`

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
diegomrsantos: I wouldn't trust recommendations to guarantee safety. I believe it is better to enforce it in the code. It is already implicitly enforced by the ImmutableSet, but I believe it is better to enforce requirements explicitly than relying on implicit checks. If this data structure is changed in the future to one that accepts null and I believe is reasonable to assume that this requirement could be neglected and then the code is unsafe.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
diegomrsantos: This class uses `Optional.ofNullable` at line 307.

diegomrsantos: ah probably it is cause Optional is serialised as null.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
diegomrsantos: makes sense.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
LukasGasior1: yup, in this case receiving null is expected, hence `ofNullable`

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
LukasGasior1: added Objects.requireNonNull check above

diegomrsantos (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBookEntry.java
diegomrsantos: We agreed to discuss this with the team.

PR #553, opened 2021-11-16
Introduce auto code formatter

diegomrsantos:
This PR introduces auto code formatter with spotless and google java format.

The main motivation is to be able to auto format code as described bellow:

```
user@machine repo % ./gradlew build
:spotlessJavaCheck FAILED
  The following files had format violations:
  src\main\java\com\diffplug\gradle\spotless\FormatExtension.java
    -\t\t····if·(targets.length·==·0)·{
    +\t\tif·(targets.length·==·0)·{
  Run './gradlew spotlessApply' to fix these violations.
user@machine repo % ./gradlew spotlessApply
:spotlessApply
BUILD SUCCESSFUL
user@machine repo % ./gradlew build
BUILD SUCCESSFUL
```


build.gradle
```diff
@@ -130,6 +131,34 @@ sonarqube {
...
+            trimTrailingWhitespace()
+            indentWithSpaces() // Takes an integer argument if you don't like 4
+            endWithNewline()
+        }
+        java {
+            // don't need to set target, it is inferred from java
+
+            // apply a specific flavor of google-java-format
+            googleJavaFormat('1.13.0').aosp().reflowLongStrings()
+
```
siy, line 153: AOSP style uses spaces for indentation. Although I'm completely in favor of this change, it will cause a lot of whitespace changes with the current style, which is TAB-based. And there might be objections from other team members regarding it. So, before introducing this change, we need to achieve a compromise across the team.

Another observations:
- AOSP uses double indentation (8 spaces) for wrapped function calls. This is fine for single wrapping without nested wraps. Long fluent call chains, with nested calls (lambdas) also wrapping, looks much less pretty. That's why we've using single indent (1 tab, equal to 4 spaces).
- We don't accept `if` without parentheses at all, AOSP allows one-line `if` statements
- We have much wider line hard wrap margin - 150 characters

siy (changes requested):

build.gradle
diegomrsantos: AOSP has been removed.

LukasGasior1 (approved):

talekhinezh (approved):

PR #554, opened 2021-11-18
Add inbound message counters

LukasGasior1:
Additional counters for how long it takes to process an inbound message.

### DIFF
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/InboundMessage.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/InboundMessage.java
@@ -66,30 +66,4 @@ package com.radixdlt.network.messaging;
 
 import com.radixdlt.network.p2p.NodeId;
 
-import java.util.Arrays;
-import java.util.Objects;
-
-public record InboundMessage(long receiveTime, NodeId source, byte[] message) {
-	@Override
-	public int hashCode() {
-		return Objects.hash(receiveTime, source, Arrays.hashCode(message));
-	}
-
-	@Override
-	public boolean equals(Object obj) {
-		if (this == obj) {
-			return true;
-		}
-		if (obj instanceof InboundMessage other) {
-			return receiveTime == other.receiveTime
-				&& Objects.equals(this.source, other.source)
-				&& Arrays.equals(this.message, other.message);
-		}
-		return false;
-	}
-
-	@Override
-	public String toString() {
-		return String.format("%s[receiveTime=%s,source=%s]", getClass().getSimpleName(), receiveTime, source);
-	}
-}
+public record InboundMessage(long receiveTime, NodeId source, byte[] message) { }
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/InboundMessage.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/InboundMessage.java
@@ -80,8 +80,7 @@ public record InboundMessage(long receiveTime, NodeId source, byte[] message) {
 		if (this == obj) {
 			return true;
 		}
-		if (obj instanceof InboundMessage) {
-			final var other = (InboundMessage) obj;
+		if (obj instanceof InboundMessage other) {
 			return receiveTime == other.receiveTime
 				&& Objects.equals(this.source, other.source)
 				&& Arrays.equals(this.message, other.message);
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/InboundMessage.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/InboundMessage.java
@@ -65,56 +65,14 @@
 package com.radixdlt.network.messaging;
 
 import com.radixdlt.network.p2p.NodeId;
-import com.radixdlt.utils.Bytes;
 
 import java.util.Arrays;
 import java.util.Objects;
 
-/**
- * A raw message received from a peer, before decoding.
- */
-public final class InboundMessage {
-	private final NodeId source;
-	private final byte[] message;
-
-	/**
-	 * Creates an inbound message with the specified source and message.
-	 *
-	 * @param source The source of the message.
-	 * @param message The message received.
-	 * @return a constructed {@code InboundMessage}
-	 */
-	public static InboundMessage of(NodeId source, byte[] message) {
-		return new InboundMessage(source, message);
-	}
-
-	private InboundMessage(NodeId source, byte[] message) {
-		// Null checking not performed for high-frequency interface
-		this.source = source;
-		this.message = message;
-	}
-
-	/**
-	 * Returns the source of the message.
-	 *
-	 * @return the source of the message.
-	 */
-	public NodeId source() {
-		return source;
-	}
-
-	/**
-	 * Returns the message.
-	 *
-	 * @return the message.
-	 */
-	public byte[] message() {
-		return message;
-	}
-
+public record InboundMessage(long receiveTime, NodeId source, byte[] message) {
 	@Override
 	public int hashCode() {
-		return Arrays.hashCode(message) * 31 + Objects.hashCode(source);
+		return Objects.hash(receiveTime, source, Arrays.hashCode(message));
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/InboundMessage.java
```diff
@@ -124,13 +82,15 @@ public boolean equals(Object obj) {
 		}
 		if (obj instanceof InboundMessage) {
```
siy, line 83: you can use pattern matching in instanceof and remove next line:

`if (obj instanceof InboundMessage other) {`

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/MessageCentralImpl.java
```diff
@@ -163,6 +169,11 @@
 	}
 
 	private Optional<MessageFromPeer<Message>> processInboundMessage(InboundMessage inboundMessage) {
+		final var messageQueuedTime = System.currentTimeMillis() - inboundMessage.receiveTime();
+		avgMessageQueuedTime.update(messageQueuedTime);
+		totalMessageQueuedTime = Math.max(totalMessageQueuedTime + messageQueuedTime, 0L);
+		updateCounters();
```
siy, line 175: Please, use `Stopwatch` to measure time.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/MessageCentralImpl.java
```diff
@@ -172,6 +183,10 @@
 					return Optional.empty();
 				},
 				messageFromPeer -> {
+					final var messageProcessingTime = System.currentTimeMillis() - processingStart;
+					avgMessageProcessingTime.update(messageProcessingTime);
+					totalMessageProcessingTime = Math.max(totalMessageProcessingTime + messageProcessingTime, 0L);
+					updateCounters();
 					if (log.isTraceEnabled()) {
```
siy, line 190: I think that lambda is long enough to deserve a dedicated method.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/PeerChannel.java
```diff
@@ -219,10 +219,12 @@ private void finalizeHandshake(AuthHandshakeResult handshakeResult) {
 	}
 
 	private void handleMessage(ByteBuf buf) throws IOException {
+		final var receiveTime = System.currentTimeMillis();
+
```
siy, line 223: Please, use `Stopwatch` to measure time.

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/messaging/InboundMessageTest.java
```diff
@@ -83,7 +83,7 @@
 
 	@Before
 	public void setUp() {
-		this.inboundMessage = InboundMessage.of(nodeId, this.message);
+		this.inboundMessage = new InboundMessage(0L, nodeId, this.message);
```
siy, line 86: What is the reason for switching from factory method to direct constructor call?

siy (changes requested):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/InboundMessage.java
LukasGasior1: nice :+1:

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/MessageCentralImpl.java
LukasGasior1: I wouldn't like to introduce a stateful property (which as far as I see Stopwatch is) to InboundMessage

LukasGasior1: I've used it for measuring the processing time below.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/MessageCentralImpl.java
LukasGasior1: I think it's easier to read to have it in one place in this case. It would seem a bit strange to have a separate method like: `logAndUpdateCountersForProcessedMessage(MessageFromPeer msg, Stopwatch processingStopwatch)`

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/messaging/InboundMessageTest.java
LukasGasior1: I'm not sure if it makes sense to keep it since we can't make record's constructor private anyway. wdyt?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/MessageCentralImpl.java
diegomrsantos: This is a good source about the topic https://www.baeldung.com/java-measure-elapsed-time

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/InboundMessage.java
```diff
@@ -65,72 +65,31 @@
...
 		}
 		return false;
 	}
 
 	@Override
 	public String toString() {
-		return String.format("%s[%s:%s]", getClass().getSimpleName(), source, Bytes.toHexString(message));
+		return String.format("%s[receiveTime=%s,source=%s]", getClass().getSimpleName(), receiveTime, source);
 	}
 }
```
diegomrsantos, line 95: These methods are generated by the compiler

LukasGasior1: yeah but the generated implementation is a bit more noisy (it includes the meaningless hashcode of message bytes)

siy: But you have the same in your hashCode implementation. Also, I really see no point to switching to records without using default hashCode and equals. After all, it was one of the motivations behind the introduction of the records.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/MessageCentralImpl.java
siy: something like `logReceivedMessage()` would work too

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/PeerChannel.java
siy, line 223: Please, use `Time` class as the source of timestamp. You may refer to `Message` class for details.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/MessageCentralImpl.java
```diff
@@ -163,6 +171,11 @@
 	}
 
 	private Optional<MessageFromPeer<Message>> processInboundMessage(InboundMessage inboundMessage) {
+		final var messageQueuedTime = System.currentTimeMillis() - inboundMessage.receiveTime();
```
siy, line 174: Please, use `Time` class as the source of the time. This will make timestamping consistent across the codebase.

siy (changes requested):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/InboundMessage.java
LukasGasior1: this needs some more investigation, I don't know why but EqualsVerifier fails for the record's generated equals/hashCode

LukasGasior1: but even with these being override, using records saves some boilerplate on getters

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/MessageCentralImpl.java
LukasGasior1: done

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/MessageCentralImpl.java
LukasGasior1: done

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/messaging/InboundMessage.java
siy: I believe that with records, we can remove test with `EqualsVerifier`. The intent of this test is to ensure that `equals()` and `hashCode()` are written correctly, but this is not necessary when the compiler is responsible for this part. So, my suggestion is to remove both, test and implementations for these methods.

LukasGasior1: ok, done

siy (approved):
LGTM

diegomrsantos (approved):

PR #569, opened 2021-12-07
Remove extra branch nibble property

michalzajda:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/tree/PMT.java
```diff
@@ -337,9 +337,9 @@ PMTExt splitExtension(PMTKey remainder, PMTNode current, PMTAcc acc) {
...
 			// XXX TODO: should this go to cache?! (now it won't as it's not added to accumulator)
 			// XXX TODO: this is probably illegal. See page 21 of the yellowpaper
-			return new PMTExt(remainder.getFirstNibble(), remainder.getTailNibbles(), current.getValue());
+			return new PMTExt(remainder.getTailNibbles(), current.getValue());
 		} else {
-			var newShorter = new PMTExt(remainder.getFirstNibble(), remainder.getTailNibbles(), current.getValue());
+			var newShorter = new PMTExt(remainder.getTailNibbles(), current.getValue());
 			acc.add(newShorter);
 			return newShorter;
 		}
```
siy, line 345: A number of too long methods. Too deeply nested code. Too long processing in switch cases. 
I think this code should be refactored to make it readable.

siy (commented):
Please, review the implementation once again and make it more readable and maintainable.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/tree/PMT.java
michalzajda: @siy  Which method is too long? 

The big case currently is for capturing correct algorithm. Once ready and with full test coverage we will refactor to maintainable form. Repetitions are actually here now to make reading easier, as the algo has complex edge cases. Also the case may benefit from java17 pattern matching

diegomrsantos (approved):

PR #586, opened 2021-12-15
Update Core API documentation

talekhinezh:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/api.yaml
```diff
@@ -652,12 +577,38 @@ tags:
       `/construction/finalize` in order to get a `signed_transaction`. This can then be submitted to
       `construction/submit`. You have successfully submitted a XRD transfer transaction!
```
dhedey, line 578: I would say `an XRD transfer`, but this is super minor :)

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/api.yaml
```diff
@@ -760,16 +720,23 @@ tags:
 
       ## Update Balance Store
 
-      We can now start updating some balance store given that we have balance changes.
+      Now that we can compute balance chaneg we can now start updating some balance store.
```
dhedey, line 723: change*

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/api.yaml
```diff
@@ -1017,11 +1044,6 @@ tags:
       the validators performing consensus. As a consequence of this, it is _not_ guaranteed to be increasing, and it
```
dhedey, line 1044: I've realised I'm missing a closing bracket in the `post_transaction_state_accumulator` line above.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/api.yaml
```diff
@@ -1017,11 +1044,6 @@ tags:
...
       may be affected by Byzantine nodes.
 
-      Not every transaction will have such a `RoundData` operation, for example, due to multiple transactions being
```
dhedey, line 1020: Why was this removed?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/api.yaml
dhedey, line 1044: I think you explained that the timestamp is actually the median - we should probably mention this. (It's certainly safer than the mean)

dhedey (commented):
Good changes, thanks. A few minor notes.

dhedey (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/api.yaml
```diff
@@ -63,40 +49,35 @@ info:
...
+    transaction by transaction.
 
-    ![Data API Flow](https://raw.githubusercontent.com/radixdlt/radixdlt/feature/open-api/radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/documentation/data_sequence_flow.png)
+    ![Data API Flow](https://raw.githubusercontent.com/radixdlt/radixdlt/feature/update-documentation/radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/documentation/data_sequence_flow.png)
 
     # Construction API Flow
 
-    Integrators can make use of the Construction API to construct and submit transactions
-    to the network.
+    The Construction API can be used to construct and submit transactions to the network.
```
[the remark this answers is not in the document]
talekhinezh: I wouldn't say it's the only reason and maybe not even the main reason. The main reason is perhaps to be able to support offline transaction creation (something we don't support now but may do so in the long term). In this case, probably not worth mentioning reasoning for some of the APIs.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/api.yaml
talekhinezh: Would rather not specify how it's derived at this time as I would like to see the implementation of this change actually in the near future.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/api.yaml
talekhinezh: It's not really explained correctly as it implies *most* transactions will have a `RoundData` operation but no user transaction will have a `RoundData` operation. I'll refactor how this is worded

PR #588, opened 2021-12-22
Feature/p2p proxy additional configs

LukasGasior1:
* add `peer_whitelist` config
* don't broadcast proxied node info


radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
```diff
@@ -145,6 +145,12 @@
    */
   Duration issuedProxyCertificateValidityDuration();
 
+  /** Specifies a list of peers that this node can connect to. If empty, all peers are allowed. */
+  ImmutableSet<NodeId> peerWhitelist();
```
dhedey, line 149: Can we call this AllowList/DenyList? See eg https://www.ncsc.gov.uk/blog-post/terminology-its-not-black-and-white  (and ditto throughout the review)

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PeerManager.java
```diff
@@ -262,6 +263,10 @@ public void tryConnect(RadixNodeUri uri) {
       return PEER_BANNED.result();
     }
 
+    if (!this.config.peerWhitelist().contains(nodeId) && !this.config.peerWhitelist().isEmpty()) {
```
dhedey, line 266: I don't really like the empty allow list meaning everything is allowed -- or at least, it might make sense at configuration time, but I don't like this leaking out beyond the config class.

I quite like the idea of either separating it into two properties: `usePeerAllowList()` and `peerAllowList()` or just exposing `peerIsAllowed(nodeId)`

radixdlt-core/radixdlt/src/main/resources/default.config
```diff
@@ -171,6 +171,15 @@ ntp.pool=pool.ntp.org
...
 # network.p2p.proxy.issued_certificate_validity_duration_ms=3600000
 
+# Specifies a list of peers that this node can connect to.
+# If empty, all peers are allowed.
+# Default: ""
+# network.p2p.peer_whitelist=
+
+# Specifies a list of peers that will not be exposed to other peers.
+# Default: ""
+# network.p2p.private_peers=
```
dhedey, line 181: For these two fields - are these meant to be a comma separated list of node addresses? Such as EG radix://rdx1qthu8yn06k75dnwpkysyl8smtwn0v4xy29auzjlcrw7vgduxvnwnst6derj@99.80.126.26 ? Or is this a node address without the IP address?

Might be worth making that clear. 

But just had a thought - if that address includes an IP address, but the validator node is on an internal network; it will have a private IP. Similarly, the sentry nodes will have both public and private IPs, but the validator node should ideally communicate with them over the private network -- this will mean the sentry nodes will have two node addresses - a private and public one.

Is this understanding correct? Will this break anything?

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/discovery/PeerDiscoveryTest.java
```diff
@@ -129,4 +143,56 @@ public void when_unexpected_response_then_ban_peer() throws Exception {
             .get()
             .isBanned());
   }
+
+  @Test
+  public void when_get_peers_then_shouldnt_return_private_peers() {
```
dhedey, line 148: Very minor, but slight preference for `should_not` here as shouldnt looks weird without the apostrophe.

radixdlt-core/radixdlt/src/main/resources/default.config
dhedey: (NB - ditto for `# network.p2p.proxy.authorized_proxied_nodes` above - but `network.p2p.proxy.authorized_proxies` specifies this better 👍 )

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/proxy/PeerProxyTest.java
```diff
@@ -236,4 +239,71 @@ public void test_peer_proxy_disabled() throws Exception {
...
+                        PROXY_NODE, CounterType.NETWORKING_ROUTING_FORWARDED_MESSAGES)
+                    == 1L);
+
+    // validator should only be connected to the proxy node
+    final var connectedPeers =
+        testNetworkRunner.peerManager(VALIDATOR_NODE).activeChannels().stream()
+            .map(PeerChannel::getRemoteNodeId)
+            .collect(Collectors.toSet());
+    assertTrue(connectedPeers.contains(nodeIdOf(PROXY_NODE)));
+    assertFalse(connectedPeers.contains(nodeIdOf(EXTERNAL_NODE)));
```
dhedey, line 299: Is it worth testing that the External Node can send a message to the Validator via the proxy?

In fact, does the External Node know about the Validator at all if it's not in its address book? Do we actually just need to hide the IP when sharing the address book contents?

dhedey (changes requested):
Nice review - a few suggestions for some changes

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
LukasGasior1: Sure, I'll rename that. I would have never thought that someone might consider this a "casually pejorative wording" but I guess that's just signum temporis of the 21st century ;)

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PeerManager.java
LukasGasior1: I think I like the idea of a separate `usePeerAllowList` property more. I'll add it.

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/proxy/PeerProxyTest.java
LukasGasior1: The external node won't have an entry for the validator node in its address book but it will have an entry for the proxy node and a corresponding proxy certificate (issued and signed by the validator node).

LukasGasior1: > Is it worth testing that the External Node can send a message to the Validator via the proxy?

I've extended the test

radixdlt-core/radixdlt/src/main/resources/default.config
LukasGasior1: The config doesn't include an IP address, updated the comments.

dhedey (approved):
LGTM after markups, cheers

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
```diff
@@ -145,119 +146,168 @@
    */
   Duration issuedProxyCertificateValidityDuration();
 
+  /** Specifies whether the peer allow list should be used. */
+  boolean usePeerAllowList();
```
diegomrsantos, line 150: Is there any use case where we would have `peerAllowList` non empty and this flag `false`? Otherwise, we could just use the former, I guess.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/discovery/PeerDiscovery.java
```diff
@@ -97,10 +98,11 @@
 public final class PeerDiscovery {
   private static final Logger log = LogManager.getLogger();
 
-  private static final int MAX_PEERS_IN_RESPONSE = 50;
+  private static final long MAX_PEERS_IN_RESPONSE = 50;
```
diegomrsantos, line 101: I would keep it as `int` and convert `MAX_PEERS_IN_RESPONSE - 1` to `long`.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
```diff
@@ -145,119 +146,168 @@
...
-      private NodeId parseNodeId(String s) {
-        try {
-          return NodeId.fromPublicKey(addressing.forNodes().parse(s));
-        } catch (DeserializeException e) {
-          throw new IllegalArgumentException("Can't parse node ID from " + s, e);
-        }
-      }
-    };
+    final var config =
+        new P2PConfig() {
```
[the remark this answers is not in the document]
LukasGasior1: already changed to a record in my next PR

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
```diff
@@ -145,119 +146,168 @@
...
+          public ImmutableList<String> seedNodes() {
+            return Arrays.stream(properties.get("network.p2p.seed_nodes", "").split(","))
+                .map(String::trim)
+                .filter(hn -> !hn.isEmpty())
+                .collect(ImmutableList.toImmutableList());
+          }
+
+          @Override
+          public int defaultPort() {
+            return properties.get("network.p2p.default_port", 30000);
```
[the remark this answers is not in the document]
LukasGasior1: you mean instead of dots? why?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
```diff
@@ -145,119 +146,168 @@
...
-      }
-    };
+    final var config =
+        new P2PConfig() {
+          @Override
+          public ImmutableList<String> seedNodes() {
+            return Arrays.stream(properties.get("network.p2p.seed_nodes", "").split(","))
+                .map(String::trim)
+                .filter(hn -> !hn.isEmpty())
+                .collect(ImmutableList.toImmutableList());
```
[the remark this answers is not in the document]
LukasGasior1: yup, it outputs an ImmutableList

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
LukasGasior1: see Davids comment above, without the additional flag the behavior when the list is empty is ambiguous

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/discovery/PeerDiscovery.java
LukasGasior1: that, on the other hand, makes intellij complain about unneeded type cast

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/discovery/PeerDiscoveryTest.java
diegomrsantos: I prefer to use **must** for something that can't happen.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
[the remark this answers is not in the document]
diegomrsantos: I mean for the integer literal, sorry, should have been more clear

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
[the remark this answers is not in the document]
diegomrsantos: I know that Java interfaces aren't clear about that, but the output is immutable too, I believe.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
diegomrsantos: Is it an issue? If the list is empty there is no peer to use anyway.

diegomrsantos: Ah if the list is empty it means everything is allowed, didn't know that.

diegomrsantos: If the default and expected behavior is to allow connection to all peers, I don't see a problem with the empty list, it doesn't mean the empty list enables this behavior, it is like that by default. On the other had, if the list isn't empty then it enables the connection only to the peers specified. I just don't like so much the idea of two configs that can be in conflict with each other.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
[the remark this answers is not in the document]
LukasGasior1: why should we use java interfaces then, if there are alternatives that are clear about that? ;)

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
LukasGasior1: I think it's safer to have them separate. It prevents allowing all peers if the list is accidentally left empty (for example if some script is used to generate the config).

diegomrsantos (approved):

PR #594, opened 2022-01-06
log4j vulernability Update

shambupujar:


diegomrsantos:
I believe it should be closed

PR #597, opened 2022-01-11
NT-60: improve handling of failed connections

siy:
Changed approach to handling of failed peer connection. 
The algorithm looks as follows:

With every failed connection attempt, peer is ignored for consequently increased time interval. Initial delay is set to 1 and then doubled with failed connection. Once the calculated interval exceeds 1hr, peer is completely removed from the address book.

This approach provides enough time for the failed node to restore (also about 1hr), but if the issue persists, the peer is removed, so there are no stall records in the address book. This should eliminate issues with failed peers, which are restored functioning with another IP address.

Known issue:
- If a peer has more than one address and one succeeds while others failed but not expired, failed addresses may remain stored. Such a "leftovers" will be cleared if peer will disconnect and then successfully reconnect after some time (longer than 1hr). Should not be an issue in practice.

### DIFF
--- a/licence-header.txt
+++ b/licence-header.txt
@@ -1,4 +1,4 @@
-/* Copyright 2021 Radix Publishing Ltd incorporated in Jersey (Channel Islands).
+/* Copyright $YEAR Radix Publishing Ltd incorporated in Jersey (Channel Islands).
  *
  * Licensed under the Radix License, Version 1.0 (the "License"); you may not use this
  * file except in compliance with the License. You may obtain a copy of the License at:
--- a/licence-header.txt
+++ b/licence-header.txt
@@ -1,4 +1,4 @@
-/* Copyright $YEAR Radix Publishing Ltd incorporated in Jersey (Channel Islands).
+/* Copyright 2021 Radix Publishing Ltd incorporated in Jersey (Channel Islands).
  *
  * Licensed under the Radix License, Version 1.0 (the "License"); you may not use this
  * file except in compliance with the License. You may obtain a copy of the License at:
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBook.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBook.java
@@ -64,8 +64,6 @@
 
 package com.radixdlt.network.p2p.addressbook;
 
-import static java.util.function.Predicate.not;
-
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.inject.Inject;
@@ -79,10 +77,10 @@ import com.radixdlt.network.p2p.addressbook.AddressBookEntry.PeerAddressEntry;
 import com.radixdlt.network.p2p.addressbook.AddressBookEntry.PeerAddressEntry.LatestConnectionStatus;
 import com.radixdlt.network.p2p.proxy.VerifiedProxyCertificate;
 import com.radixdlt.utils.Pair;
+
+import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.Comparator;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
@@ -90,67 +88,30 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Stream;
 
+import static java.util.function.Predicate.not;
+
 /** Manages known peers network addresses and their metadata. */
 public final class AddressBook {
-
-  /**
-   * A stateful comparator for known peer addresses that uses both their latest connection status
-   * (persistent) and a number of failed connection attempts (volatile). Failure counts are used to
-   * cycle the addresses for retries. The entries are sorted in the following manner: successful ->
-   * unknown (no connection attempted yet) -> failed (by num of failures)
-   */
-  private static final class PeerAddressEntryComparator implements Comparator<PeerAddressEntry> {
-    private final Map<RadixNodeUri, Integer> failureCounts = new HashMap<>();
-
-    private int toIntValue(PeerAddressEntry peerAddressEntry) {
-      return peerAddressEntry
-          .getLatestConnectionStatus()
-          .map(
-              latestConnectionStatus -> {
-                if (latestConnectionStatus == LatestConnectionStatus.SUCCESS) {
-                  return 1;
-                } else {
-                  return -(1 + failureCounts.getOrDefault(peerAddressEntry.getUri(), 0));
-                }
-              })
-          .orElse(0);
-    }
-
-    @Override
-    public int compare(PeerAddressEntry a, PeerAddressEntry b) {
-      return Integer.compare(toIntValue(b), toIntValue(a));
-    }
-
-    void incFailures(RadixNodeUri uri) {
-      synchronized (this.failureCounts) {
-        final var curr = this.failureCounts.getOrDefault(uri, 0);
-        this.failureCounts.put(uri, curr + 1);
-      }
-    }
-
-    void resetFailures(RadixNodeUri uri) {
-      synchronized (this.failureCounts) {
-        this.failureCounts.remove(uri);
-      }
-    }
-  }
+  private static final Duration INITIAL_POSTPONE = Duration.ofMinutes(5);
+  private static final Duration MAX_POSTPONE = Duration.ofHours(1);
 
   private final RadixNodeUri self;
   private final EventDispatcher<PeerEvent> peerEventDispatcher;
   private final AddressBookPersistence persistence;
+  private final PeerPostponingManager postponingManager;
   private final Object lock = new Object();
   private final Map<NodeId, AddressBookEntry> knownPeers = new ConcurrentHashMap<>();
-  private final PeerAddressEntryComparator addressEntryComparator =
-      new PeerAddressEntryComparator();
 
   @Inject
   public AddressBook(
       @Self RadixNodeUri self,
       EventDispatcher<PeerEvent> peerEventDispatcher,
-      AddressBookPersistence persistence) {
+      AddressBookPersistence persistence,
+      Clock clock) {
     this.self = Objects.requireNonNull(self);
     this.peerEventDispatcher = Objects.requireNonNull(peerEventDispatcher);
     this.persistence = Objects.requireNonNull(persistence);
+    this.postponingManager = new PeerPostponingManager(clock, INITIAL_POSTPONE, MAX_POSTPONE);
     persistence.getAllEntries().forEach(e -> knownPeers.put(e.getNodeId(), e));
     cleanup();
   }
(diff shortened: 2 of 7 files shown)


licence-header.txt
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2021 Radix Publishing Ltd incorporated in Jersey (Channel Islands).
+/* Copyright $YEAR Radix Publishing Ltd incorporated in Jersey (Channel Islands).
```
diegomrsantos, line 1: This is not necessary. I asked @russellharvey and he said this should always be 2021.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBook.java
```diff
@@ -227,20 +194,28 @@ private boolean sameNetworkHrp(RadixNodeUri uri) {
...
+
+    this.postponingManager.recordSuccess(radixNodeUri);
   }
 
   public void addOrUpdatePeerWithFailedConnection(RadixNodeUri radixNodeUri) {
     this.addOrUpdatePeerWithLatestConnectionStatus(radixNodeUri, LatestConnectionStatus.FAILURE);
-    this.addressEntryComparator.incFailures(radixNodeUri);
+
+    if (!this.postponingManager.recordFailure(radixNodeUri)) {
+      synchronized (lock) {
```
diegomrsantos, line 214: Is this lock necessary? `knownPeers` is thread-safe. Not sure bout `persistence` tho.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBook.java
```diff
@@ -92,65 +91,33 @@
...
-      }
-    }
-  }
+  private static final Duration INITIAL_POSTPONE = Duration.ofMinutes(5);
+  private static final Duration MAX_POSTPONE = Duration.ofHours(1);
 
   private final RadixNodeUri self;
   private final EventDispatcher<PeerEvent> peerEventDispatcher;
   private final AddressBookPersistence persistence;
+  private final PeerPostponingManager postponingManager;
```
diegomrsantos, line 100: Not sure about this name, but can't think of a better one. @dhedey what do you think?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/PeerPostponingManager.java
```diff
@@ -0,0 +1,170 @@
...
+  private final class PostponeInfo {
+    private final Instant firstFailure;
+    private Duration postponeDuration;
+
+    PostponeInfo() {
+      this.firstFailure = now();
+      this.postponeDuration = initialPostpone;
+    }
+
+    boolean survivesFailure() {
```
dhedey, line 153: This function has a side effect which was unexpected - I'd rather the side effect be clear from the method name (rather than the return statement) - or have separate methods: `void recordConnectionFailure()` and `bool shouldBeRemovedFromAddressBook()`

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/addressbook/PeerPostponingManagerTest.java
```diff
@@ -0,0 +1,155 @@
...
+    manager.recordFailure(peer2);
+
+    // Both addresses are postponed now
+    assertTrue(manager.shouldIgnore(peer1));
+    assertTrue(manager.shouldIgnore(peer2));
+
+    manager.recordSuccess(peer1);
+
+    // Both addresses are available now
+    assertFalse(manager.shouldIgnore(peer1));
```
dhedey, line 127: I'd rather the clock is explicitly set to be passed the 1 minute mark, rather than relying on the number of internal calls to `clock.instant()` to ensure the clock is sufficiently advanced to allow this.

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/addressbook/AddressBookTest.java
```diff
@@ -136,51 +139,54 @@ public void address_book_should_sort_entries_by_latest_connection_status() {
...
+    System.out.println(bestAddr5);
+    assertTrue(bestAddr5 == addr3 || bestAddr5 == addr4);
   }
 
   @Test
-  public void address_book_should_cycle_failed_connection_uris() {
-    final var self =
-        RadixNodeUri.fromPubKeyAndAddress(
-            1, ECKeyPair.generateNew().getPublicKey(), "127.0.0.10", 30303);
+  public void address_book_should_not_show_failed_connection_uris_while_they_postponed() {
```
dhedey, line 147: Very nit-picky, but slight typo: `while_they_are_postponed`*

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/addressbook/AddressBookTest.java
```diff
@@ -103,18 +109,15 @@ public void address_book_should_filter_out_peers_with_different_network_hrp() {
...
-    final var self =
-        RadixNodeUri.fromPubKeyAndAddress(
-            1, ECKeyPair.generateNew().getPublicKey(), "127.0.0.10", 30303);
     final var peerKey = ECKeyPair.generateNew().getPublicKey();
     final var peerId = NodeId.fromPublicKey(peerKey);
-    final var addr1 = RadixNodeUri.fromPubKeyAndAddress(1, peerKey, "127.0.0.1", 30303);
-    final var addr2 = RadixNodeUri.fromPubKeyAndAddress(1, peerKey, "127.0.0.2", 30303);
-    final var addr3 = RadixNodeUri.fromPubKeyAndAddress(1, peerKey, "127.0.0.3", 30303);
-    final var addr4 = RadixNodeUri.fromPubKeyAndAddress(1, peerKey, "127.0.0.4", 30303);
+    final var self = createNodeUri(1, "127.0.0.10");
```
dhedey, line 114: I like the createNodeUri helper method

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/addressbook/AddressBookTest.java
```diff
@@ -136,51 +139,54 @@ public void address_book_should_sort_entries_by_latest_connection_status() {
 
     sut.addOrUpdatePeerWithSuccessfulConnection(addr4);
     final var bestAddr5 = sut.findBestKnownAddressById(peerId).orElseThrow();
-    assertEquals(addr4, bestAddr5);
+    System.out.println(bestAddr5);
```
dhedey, line 142: Was this left in unintentionally?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/PeerPostponingManager.java
```diff
@@ -0,0 +1,170 @@
...
+  private final Duration initialPostpone;
+  private final Duration maxPostpone;
+
+  public PeerPostponingManager(Clock clock, Duration initialPostpone, Duration maxPostpone) {
+    this.clock = clock;
+    this.initialPostpone = initialPostpone;
+    this.maxPostpone = maxPostpone;
+  }
+
+  public boolean recordFailure(RadixNodeUri uri) {
```
dhedey, line 95: Ditto here - we have a method here with a side effect and a returned boolean. How would you feel about separating the side effect (record failure) from the question (shouldRemoveFromAddressBookDueToFailures?)

dhedey (commented):
Looks good. Some minor naming suggestions.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBook.java
siy: Persistence layers is not thread safe

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/PeerPostponingManager.java
siy: Here I had a choice between action/query separation and creation of potential space for race condition. Given how usually hard to nail down and fix race conditions, I prefer to live with atomic action-query. But we may discuss naming :)

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/PeerPostponingManager.java
siy: This component needs to maintain some data which change over the time. 
Usually, there are two approaches - one you mention above, but it requires independent maintenance of the data, for example using dedicated thread (or similar concept, for example, time-based notifications). 
Another approach is to update time-based data when data are accessed. This method is the single point, which can be used for updates of internal state conveniently.

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/addressbook/AddressBookTest.java
siy: Good catch. Cleaned it up.

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/addressbook/AddressBookTest.java
siy: Fixed

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/addressbook/PeerPostponingManagerTest.java
siy: Reworked to use special kind of Clock.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBook.java
diegomrsantos: would it be too hard to do that?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/PeerPostponingManager.java
dhedey: I don't necessarily see a problem with lack of atomicity here, as long as `shouldRemoveFromAddressBookDueToFailures` doesn't make any assumption that the node exists in `postponedPeers`. Indeed, if a second successful request comes in whilst processing a first unsuccessful request, it'd be better if the first request _didn't_ remove the address from the address book! aha. But it's probably quite unlikely.

Alternatively, I'm happy to keep the current structure - but in terms of naming, my personal preferences (from imo most honest to least honest) are 1. Naming both the side effect and return `recordFailureAndReturnIfShouldBeKeptInAddressBook`, or 2. Naming just the side effect (if you have to) - in which case `recordFailure` is okay, before 3. Naming just the return (as it's easy to miss the side effect).

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/PeerPostponingManager.java
dhedey: I'm happy for a single method, but as per the comment above, I'd personally prefer the name to capture the side effect instead of the return value (`recordFailure` or `doublePostponeDuration`), or even more clearly/honestly, capture both: `doublePostponeDurationAndReturnIfShouldStayInAddressBook`

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/addressbook/PeerPostponingManagerTest.java
dhedey: Really nice! 👍

dhedey (approved):
LGTM! Couple of comments on naming.

PR #605, opened 2022-01-14
Filter self from p2p discovery if proxy is enabled

LukasGasior1:
* Filter self from p2p discovery if proxy is enabled
* add use_proxies flag
* fix dependency injection


radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
```diff
@@ -308,6 +315,14 @@ private NodeId parseNodeId(String s) {
           "peerAllowList is non empty but will be ignored because usePeerAllowList is not set");
     }
 
+    if (config.useProxies() && config.authorizedProxies().isEmpty()) {
+      throw new IllegalArgumentException("authorizedProxies can't be empty if useProxies is true");
+    }
+
+    if (!config.useProxies() && !config.authorizedProxies().isEmpty()) {
+      throw new IllegalArgumentException("authorizedProxies can't be set if useProxies is false");
```
siy, line 323: While I understand the intent of this check, I think it might be inconvenient in practice. 
For example, if a user had configured proxies and then decided to just disable them (for example, experimenting or checking something), he will ought to clear the configuration. But presence of the proxies if they are disabled should not prevent the node from start.

siy (approved):
LGTM with one minor comment

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
LukasGasior1: yeah, I think you're right. I'll change that to just log a warning

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
```diff
@@ -308,6 +315,14 @@ private NodeId parseNodeId(String s) {
           "peerAllowList is non empty but will be ignored because usePeerAllowList is not set");
     }
 
+    if (config.useProxies() && config.authorizedProxies().isEmpty()) {
+      throw new IllegalArgumentException("authorizedProxies can't be empty if useProxies is true");
+    }
+
+    if (!config.useProxies() && !config.authorizedProxies().isEmpty()) {
+      log.warn("authorizedProxies config will be ignored because useProxies is false");
+    }
```
siy, line 324: Super!

siy (approved):
LGTM

PR #620, opened 2022-01-26
WIP Peers view for proxied peers

siy:


siy:
Requires significant rework

PR #624, opened 2022-01-31
Next round of code smell fixes

siy:
A bunch of minor cleanups to address Sonar code smells.


radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
```diff
@@ -183,7 +183,8 @@ public EpochManager(
     this.syncErrorResponseProcessors =
         isValidator ? Set.of(initialBFTSync.errorResponseProcessor()) : Set.of();
     this.bftUpdateProcessors =
-        isValidator
+        isValidator // FIXME: initialBFTSync::processBFTUpdate does nothing, do we actually need it
```
dhedey, line 186: I'm concerned about increasing the number of untracked comments like this in the code base. Can we do one of:
* Investigate this now so we don't need to leave a comment, if you feel it fits with the current work you're doing
* Raise a ticket for this and mention the ticket number in the code comment

What are your thoughts?

dhedey (changes requested):
Lots of nice tweaks / neatening! One comment about the code comment.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
siy: It requires somewhat deeper look before fix. This is out of scope of the current task.
I believe there is a dedicated ticket to transform all TODO's into tickets.

dhedey: I'm happy if we raise another ticket. But the best time to create that ticket is now whilst you have the context, to help whoever picks it up know where to start. Would you be happy to raise it?

siy: Actually, "FIXME" appeared exactly because I had no full context. Anyway, I've decided to look deeper and resolved the issue.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/DispatcherModule.java
```diff
@@ -162,12 +160,6 @@ public void configure() {
     bind(new TypeLiteral<EventDispatcher<MempoolRelayTrigger>>() {})
         .toProvider(Dispatchers.dispatcherProvider(MempoolRelayTrigger.class))
         .in(Scopes.SINGLETON);
-    bind(new TypeLiteral<EventDispatcher<MempoolFillerUpdate>>() {})
```
LukasGasior1, line 165: what's the reason to move these out of DispatcherModule?

dhedey (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/DispatcherModule.java
siy: The `MempoolFiller` and related classes are used exclusively for testing purposes. They are not configured nor used in the production code.

PR #625, opened 2022-01-31
Fix deb4docker build

siy:
Fix Java compiler option setting (include not only subprojects, but top level projects too).


Theoklitos (approved):
lgtm

PR #634, opened 2022-02-07
Add config to restrict p2p connections to certain subnets

siy:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
```diff
@@ -202,6 +204,14 @@ public static P2PConfig fromRuntimeProperties(
             .map(n -> P2PConfig.parseNodeId(addressing, n))
             .collect(ImmutableSet.toImmutableSet());
 
+    final var peerHostAllowList =
```
LukasGasior1, line 207: rename to allowed subnets?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/PeerChannel.java
```diff
@@ -259,15 +259,41 @@ private void handleMessage(ByteBuf buf) throws IOException {
...
+            .anyMatch(subnet -> subnet.matches(remoteAddress.getAddress()));
+
+    if (!allowedIp) {
+      // Switch to dormant state to prevent starting initialization sequence
+      this.state = ChannelState.REJECTED;
+
+      log.warn(
+          "Remote address {} does not belong to allowed subnets, scheduling closing connection",
+          remoteAddress);
+      ctx.close()
```
LukasGasior1, line 290: can we just reuse `this.disconnect()`?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/PeerChannel.java
```diff
@@ -259,15 +259,41 @@ private void handleMessage(ByteBuf buf) throws IOException {
...
   public void channelActive(ChannelHandlerContext ctx) {
     // if we weren't able to determine peer's address earlier, it should be available now
+    var peerAddress = this.nettyChannel.remoteAddress();
+
     if (this.remoteAddress.isEmpty()) {
-      this.remoteAddress = Optional.ofNullable(this.nettyChannel.remoteAddress());
+      this.remoteAddress = Optional.ofNullable(peerAddress);
+    }
+
+    if (peerAddress != null) {
```
LukasGasior1, line 268: What if it's null? I think we might want to close the channel in that case anyway.
Something like this perhaps?:
```
if (peerAddress == null) {
  log("invalid address");
  this.disconnect();
} else if (!isSubnetAllowed(peerAddress)) {
  log("not allowed");
  this.disconnect();
} else if (this.state == ChannelState.INACTIVE) {
  this.init()
}
```

with that, we could drop the extra `REJECTED` state.

LukasGasior1 (commented):
I think we might also want to add the subnet check in `PeerOutboundBootstrapImpl` so that we don't even attempt an outgoing connection to a forbidden address.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/PeerChannel.java
siy: OK, makes sense

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/P2PConfig.java
siy: Done

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/PeerChannel.java
siy: Done

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/PeerChannel.java
```diff
@@ -259,15 +259,41 @@ private void handleMessage(ByteBuf buf) throws IOException {
...
 
+  private void checkRemoteIp(ChannelHandlerContext ctx, InetSocketAddress remoteAddress) {
+    // Check if remote belongs to allowed subnets
+    var allowedIp =
+        config.allowedSubnets().stream()
+            .anyMatch(subnet -> subnet.matches(remoteAddress.getAddress()));
+
+    if (!allowedIp) {
+      // Switch to dormant state to prevent starting initialization sequence
+      this.state = ChannelState.REJECTED;
```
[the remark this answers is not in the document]
siy: Fixed with change proposed below.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/PeerChannel.java
```diff
@@ -259,15 +258,25 @@ private void handleMessage(ByteBuf buf) throws IOException {
   @Override
   public void channelActive(ChannelHandlerContext ctx) {
     // if we weren't able to determine peer's address earlier, it should be available now
-    if (this.remoteAddress.isEmpty()) {
```
LukasGasior1, line 262: This is a bit unintuitive, but the check was there for a reason. If the node runs behind a reverse proxy (nginx), it gets peer's real ip address through a so called "proxy protocol", which is then supplied in a constructor.
Without the if statement, this would override that address with the address of Nginx.

LukasGasior1: this could probably be refactored to make it more visible

siy: Good catch. Fixed and made visible.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/Subnet.java
```diff
@@ -0,0 +1,125 @@
...
+public record Subnet(InetAddress address, int numBits) {
+  private static final Pattern IPV4 =
+      Pattern.compile(
+          "^(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\."
+              + "(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\."
+              + "(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\."
+              + "(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])$");
+  private static final Pattern MASK = Pattern.compile("^[0-9]{1,2}$");
+
+  public static Optional<Subnet> fromString(String subnet) {
```
diegomrsantos, line 84: Why this method returns option? Maybe I'm missing something here, but I think it should thrown an exception if the String can't be parsed.

LukasGasior1 (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/Subnet.java
siy: Hate exceptions, but here it's justified and makes behavior consistent with other configuration settings.
Added method which throws an exception.

PR #636, opened 2022-02-11
Remove maximum message length size hardcoding

siy:
This PR adds message length as a part of RE configuration and removes hardcoded limits.
Network configurations are updated to reflect this change, but actual limit (i.e. maximal message length which is available to users) remains unchanged.


radixdlt-core/radixdlt/src/test/java/com/radixdlt/statecomputer/radixengine/AbstractRadixEngineTest.java
```diff
@@ -0,0 +1,166 @@
...
+import com.radixdlt.store.berkeley.BerkeleyAdditionalStore;
+import com.radixdlt.utils.PrivateKeys;
+import com.radixdlt.utils.UInt256;
+import java.util.Map;
+import java.util.Set;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.rules.TemporaryFolder;
+
+public abstract class AbstractRadixEngineTest {
```
dhedey, line 108: I slightly prefer naming `_Base` instead if `Abstract_` (eg `RadixEngineTestBase`) because it the latter sounds like a test of an AbstractRadixEngine.

radixdlt-core/radixdlt/src/test/java/com/radixdlt/statecomputer/radixengine/AbstractRadixEngineTest.java
```diff
@@ -0,0 +1,166 @@
...
+        Guice.createInjector(
+            MempoolConfig.asModule(mempoolMaxSize, 10),
+            new MainnetForkConfigsModule(),
+            new RadixEngineForksLatestOnlyModule(
+                RERulesConfig.testingDefault()
+                    .overrideFeeTable(
+                        FeeTable.create(
+                            Amount.ofSubunits(UInt256.ONE),
+                            Map.of(ValidatorRegisteredCopy.class, Amount.ofSubunits(UInt256.ONE))))
+                    .overrideMaxMessageLen(511)),
```
dhedey, line 134: For a class called a generic name such as `AbstractRadixEngineTest`, I find this override unexpected.

Is it worth instead having the class take some optional overrides in the constructor, to make the values of the important parameters in subclasses explicit in those classes?

radixdlt-engine/src/main/java/com/radixdlt/atom/TxLowLevelBuilder.java
```diff
@@ -159,37 +145,34 @@ public void instruction(REInstruction.REMicroOp op, byte[] data) {
...
+  public TxLowLevelBuilder message(byte[] bytes, int limit) throws MessageTooLongException {
+    if (bytes.length > limit) {
+      throw new MessageTooLongException(limit, bytes.length);
     }
 
-    instruction(REInstruction.REMicroOp.MSG, varLengthData(bytes));
-    return this;
-  }
+    var buf = ByteBuffer.allocate(Short.BYTES + bytes.length);
+    buf.putShort((short) bytes.length);
```
dhedey, line 154: Do we need to protect against the possibility someone ups the limit above 32,767 (max short)?

Perhaps we artificially cap the limit to max short at the top of this method?

radixdlt-engine/src/main/java/com/radixdlt/atom/TxLowLevelBuilder.java
```diff
@@ -271,14 +253,17 @@ public TxLowLevelBuilder read(SubstateId substateId) {
...
-    var particle = localUpParticles.remove(index);
-    if (particle == null) {
-      throw new IllegalStateException("Local particle does not exist: " + index);
-    }
+    validateParticle(localUpParticles.remove(index), index);
     instruction(REInstruction.REMicroOp.LDOWN, Shorts.toByteArray((short) index));
     return this;
   }
 
+  private void validateParticle(LocalSubstate particle, int index) {
```
dhedey, line 261: While we're moving this, can we update the naming? (IE this should be `validateLocalSubstate`)

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/REInstruction.java
```diff
@@ -218,7 +218,9 @@ Object read(REParser.ParserState parserState, ByteBuffer b) throws DeserializeEx
         return REFieldSerialization.deserializeSignature(b);
       }
     },
-    MSG((byte) 0xc, REOp.MSG, LengthType.VARIABLE, 1, 255) {
+    // There is no way to put actual limit here, so it is just lifted to maximum possible
+    // value, given used encoding format.
+    MSG((byte) 0xc, REOp.MSG, LengthType.VARIABLE, 1, 65535) {
```
dhedey, line 223: I think short is signed in Java, so we're restricted to half this

radixdlt-engine/src/main/java/com/radixdlt/engine/parser/REParser.java
```diff
@@ -275,4 +262,23 @@ public ParsedTxn parse(Txn txn) throws TxnParseException {
         payloadHashAndSig,
         parserState.disableResourceAllocAndDestroy);
   }
+
+  private Pair<HashCode, ECDSASignature> calculatePayloadHashAndSign(
```
dhedey, line 266: Because `Sign` looks like a verb, I think a slightly clearer name is `calculatePayloadHashAndSignature`

dhedey (approved):
Looks good, thanks for all the tidying up too! 😄

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/forks/RERulesConfig.java
```diff
@@ -253,6 +201,7 @@ public RERulesConfig overrideMaxRounds(long maxRounds) {
         this.unstakingEpochDelay,
         this.rewardsPerProposal,
         this.minimumCompletedProposalsPercentage,
-        this.maxValidators);
+        this.maxValidators,
+        this.maxMessageLen);
```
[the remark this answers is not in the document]
siy: I think that creation of test data and `overrideXXX` methods should be moved out of the class declaration into test utilities. Unfortunately, such a refactoring is quite complicated now, as there are cross-module dependencies and coupling between main code and integration tests. 

Created ticket to address this issue: https://radixdlt.atlassian.net/browse/NT-259

radixdlt-core/radixdlt/src/test/java/com/radixdlt/statecomputer/radixengine/AbstractRadixEngineTest.java
siy: Using `Abstract` prefix is a quite traditional approach for the Java classes naming, especially for classes declared as abstract (which is the case).

radixdlt-core/radixdlt/src/test/java/com/radixdlt/statecomputer/radixengine/AbstractRadixEngineTest.java
siy: Actually, this is the same issue as one mentioned above - `RERulesConfig` has factory and `overrideXXX` methods which are used exclusively in tests. To address this issue, I've created a ticket (see above).

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/REInstruction.java
siy: This limit does not play a significant role, but you're right, there are no reason to set it higher than we can actually write.

radixdlt-engine/src/main/java/com/radixdlt/engine/parser/REParser.java
siy: Actually, even this name does not reflect the real purpose of this method. Reworked it a bit and fixed naming.

radixdlt-core/radixdlt/src/test/java/com/radixdlt/statecomputer/radixengine/AbstractRadixEngineTest.java
dhedey: I've seen both conventions, but in this code base, we have two `_Base` and one `Abstract_` in the codebase already. I feel `_Base` is clearer about the intention of the class, and as well as the unclear semantics in a Test class (as per my previous comment), I also feel Abstract is a slightly weird abstraction leakage - it describes a property of the class, instead of what it's used for (and reminds me of the AbstractComponentFactoryInstantiatorBean java joke xD), but I'm ultimately happy with either.

radixdlt-core/radixdlt/src/test/java/com/radixdlt/statecomputer/radixengine/AbstractRadixEngineTest.java
dhedey: My point is that, in a test base class, I don't feel like we should have unexpected/arbitrary overrides of defaults unless clearly required / specified by a subclass. Otherwise the test set-up is hidden/implicit instead of explicit, resulting in a more fragile and harder to understand test.

radixdlt-engine/src/main/java/com/radixdlt/atom/TxLowLevelBuilder.java
dhedey: Do we ever ensure the limit is <= 32,767?

siy: No, there are no checks for that. We don't have it because it should be part of the validation of `RERulesConfig`, which is missing. Given that `RERulesConfig` instances are always hardcoded, I'm uncertain if we should have such a validation in the first place.

radixdlt-core/radixdlt/src/test/java/com/radixdlt/statecomputer/radixengine/AbstractRadixEngineTest.java
siy: I've changed the test setup to make this limit coming directly from the particular test.

radixdlt-engine/src/main/java/com/radixdlt/engine/RadixEngine.java
```diff
@@ -435,30 +438,33 @@ private TxBuilder construct(TxBuilderExecutable executable, Set<SubstateId> avoi
...
+          if (request.isDisableResourceAllocAndDestroy()) {
+            txBuilder.toLowLevelBuilder().disableResourceAllocAndDestroy();
+          }
+          for (var action : request.getActions()) {
+            this.actionConstructors.construct(action, txBuilder);
+          }
+
+          var msg = request.getMsg();
+          if (msg.isPresent()) {
+            txBuilder.message(msg.get(), maxMessageLen);
```
talekhinezh, line 464: This seems like antipattern here, you already know messageLen so why pass this in? Instead, should pass maxMessageLen into `TxBuilder` constructor

radixdlt-engine/src/main/java/com/radixdlt/engine/RadixEngine.java
```diff
@@ -517,7 +523,7 @@ private TxBuilder constructWithFees(TxnConstructionRequest request, REAddr feePa
               }
               var msg = request.getMsg();
               if (msg.isPresent()) {
-                txBuilder.message(msg.get());
+                txBuilder.message(msg.get(), maxMessageLen);
```
talekhinezh, line 526: Same comment as above

talekhinezh (changes requested):

radixdlt-core/radixdlt/src/test/java/com/radixdlt/statecomputer/radixengine/AbstractRadixEngineTest.java
dhedey: For me, I get `RadixApiBase` and `LatencyProviderBase` vs `AbstractHttpServer` (which isn't even abstract and no longer exists, I only have an old class file)

radixdlt-engine/src/main/java/com/radixdlt/atom/TxLowLevelBuilder.java
dhedey: I'm comfortable with this for now, assuming this would throw at the cast?

radixdlt-core/radixdlt/src/test/java/com/radixdlt/statecomputer/radixengine/AbstractRadixEngineTest.java
siy:
```
siy@bobryzen:~/Projects/Radix-quick-look/radixdlt$ find . -name "*Base*"
./radixdlt-java-common/build/classes/java/main/com/radixdlt/utils/Base58.class
./radixdlt-java-common/src/main/java/com/radixdlt/utils/Base58.java
./radixdlt-core/radixdlt/build/classes/java/main/com/radixdlt/qualifier/LatencyProviderBase.class
./radixdlt-core/radixdlt/src/main/java/com/radixdlt/qualifier/LatencyProviderBase.java
```
The `LatencyProviderBase` is an annotation and does not serve as a base class for other classes.

radixdlt-engine/src/main/java/com/radixdlt/atom/TxLowLevelBuilder.java
siy: No, it wouldn't. Added dedicated check to ensure length never exceed "hardware" limit.

radixdlt-engine/src/main/java/com/radixdlt/engine/RadixEngine.java
siy: Fixed

talekhinezh (changes requested):

radixdlt-engine/src/main/java/com/radixdlt/constraintmachine/REInstruction.java
```diff
@@ -218,7 +218,9 @@ Object read(REParser.ParserState parserState, ByteBuffer b) throws DeserializeEx
         return REFieldSerialization.deserializeSignature(b);
       }
     },
-    MSG((byte) 0xc, REOp.MSG, LengthType.VARIABLE, 1, 255) {
+    // There is no way to put actual limit here, so it is just lifted to maximum possible
+    // value, given used encoding format.
```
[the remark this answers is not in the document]
siy: This is already fixed

radixdlt-engine/src/main/java/com/radixdlt/atom/TxLowLevelBuilder.java
```diff
@@ -271,14 +259,17 @@ public TxLowLevelBuilder read(SubstateId substateId) {
...
-    var particle = localUpParticles.remove(index);
-    if (particle == null) {
-      throw new IllegalStateException("Local particle does not exist: " + index);
-    }
+    validateSubstate(localUpParticles.remove(index), index);
     instruction(REInstruction.REMicroOp.LDOWN, Shorts.toByteArray((short) index));
     return this;
   }
 
+  private void validateSubstate(LocalSubstate substate, int index) {
```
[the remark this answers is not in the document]
siy: The method is used in two places (not just  with `remove()`, but with `get()` too) and index is necessary exclusively to generate a useful exception message.

radixdlt-engine/src/main/java/com/radixdlt/atom/TxLowLevelBuilder.java
```diff
@@ -159,37 +147,38 @@ public void instruction(REInstruction.REMicroOp op, byte[] data) {
     }
   }
 
-  public TxLowLevelBuilder message(byte[] bytes) throws MessageTooLongException {
-    if (bytes.length > 255) {
-      throw new MessageTooLongException(bytes.length);
+  TxLowLevelBuilder message(byte[] bytes, int limit) throws MessageTooLongException {
```
talekhinezh, line 150: This also looks like antipattern, should just include `limit` in TxLowLevelBuilder constructor

talekhinezh: It is a required coupling though as it related to what's allowed to be built. It shouldn't be the caller who makes the decision on the limit. And if the caller is the decider then why can't the caller make the check on its own.

radixdlt-engine/src/main/java/com/radixdlt/atom/TxLowLevelBuilder.java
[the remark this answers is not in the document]
dhedey: How about making it instead a `validateLocalUpSubstateExists(index)`, which is called as a precondition before the get/remove on the next line? Whilst we're at it, could we rename `localUpParticles` as `localUpSubstates`?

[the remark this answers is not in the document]
siy: I think it affects existing behavior because there will be two instead of one accesses to the `localUpSubstates` map.
Also, I've made `validateSubstate` static and renamed to `validateSubstatePresentAtIndex` and think that this should address remaining objections.

talekhinezh (approved):

PR #637, opened 2022-02-11
Add missing Prometheus metrics for validator BFT stats

siy:


LukasGasior1 (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/system/prometheus/PrometheusService.java
```diff
@@ -243,6 +247,27 @@ private void appendField(StringBuilder builder, String name, Object value) {
 
   private void exportCounters(StringBuilder builder) {
     EXPORT_LIST.forEach(counterType -> generateCounterEntry(counterType, builder));
+
+    getProposalStats()
+        .ifPresent(
+            uptime -> {
+              appendCounter(
+                  builder,
+                  COUNTER_PREFIX + "radix_engine_cur_epoch_completed_proposals",
```
diegomrsantos, line 256: maybe it is better to extract it to a static constant?

siy: There is just one location, where this constant is used, so it does not make code more readable or more convenient for support and maintenance.

diegomrsantos: will this concatenation happen only once or every time this value is updated?

siy: Compiler is smart enough to optimize it.

PR #642, opened 2022-02-17
Add size property to the tree

diegomrsantos:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/tree/PMTBranch.java
```diff
@@ -156,11 +156,15 @@ public void removeNode(
       PMTKey key, PMTAcc acc, Function<PMTNode, byte[]> represent, Function<byte[], PMTNode> read) {
     final PMTPath commonPath = PMTPath.findCommonPath(this.getKey(), key);
     if (commonPath.whichRemainderIsLeft() == PMTPath.RemainingSubtree.NONE) {
-      acc.remove(this);
-      var newBranch = new PMTBranch(this);
-      newBranch.setValue(null);
-      acc.add(newBranch);
-      acc.setTip(newBranch);
+      if (branchDoesNotHaveValue()) {
```
dhedey, line 159: I'm not sure I can easily follow this algorithm just from the code review - so I'll trust this is correct :)

As this is complicated stuff, I'd suggest if you know what's going on, adding a few comments to explain the key steps of the algorithm, to help future you / other people to follow the flow and/or ramp up. The comments should focus on what we know about the state and why things are being done, not what is done, as that just duplicates the code (eg the `// remove child` comment below doesn't add value beyond the code)

Alternatively / preferentially, the conditions can be extracted to booleans and named to help the code read like prose - eg `var nodeBranchesOffAtBranchRoot = commonPath.whichRemainderIsLeft() == PMTPath.RemainingSubtree.NONE`. Although I'm still unclear what `branchDoesNotHaveValue` really means - :) - eg when would a branch have a value? What does setting acc to NotFound mean?

dhedey (approved):
This change looks good - I'll save some general thoughts for the review into main.

I take it some of the changes to PMTBranch were the bug fix? Just a thought for future, it might be nice to separate the commits for the implementation and the fix to keep the commit history clean. Not a big deal though! 👍

diegomrsantos:
> 

I totally agree. I think I couldn't easily isolate the bug fix changes in this case, but I'll be more aware of that next time.

diegomrsantos:
> 

I totally agree. I think I couldn't easily isolate the bug fix changes in this case, but I'll be more aware of that next time.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/tree/PMTBranch.java
michalzajda: What's the best form to elaborate on the algorithm? I am happy to answer, but it's just an example of many similar cases.

dhedey: It's a hard question to answer in general, but for me, the key first step to readable code is about ensuring your abstractions are consistent. Typically this will mean naming variables/methods in terms of intention, instead of implementation.

For algorithms like this, I'd be tempted to break long methods into sub methods, or at least use variables which can be clearly named, or if that's not possible, add some explanatory text at key stages of an algorithm.

Happy to run this through in a pairing session if it helps :)

diegomrsantos: > Although I'm still unclear what branchDoesNotHaveValue really means - :) - eg when would a branch have a value?

This is part of the PMT algorithm. Probably this kind of thing should be defined with a specification. Not sure if it is feasible to describe it with comments in the code.

PR #645, opened 2022-02-28
New fork vote format; add epochsBeforeEnacted and maxEpoch params

LukasGasior1:


build.gradle
```diff
@@ -157,6 +157,7 @@ allprojects {
             // through git history (see "license" section below)
             licenseHeaderFile("${project.rootDir}/licence-header.txt")
 
+            toggleOffOn()
```
diegomrsantos, line 160: Why this is necessary?

LukasGasior1: @Theoklitos ^

LukasGasior1: I use it in one test, spotless formatting doesn't really work well for test data params and makes things harder to read

diegomrsantos: I understand the motivation, but I'm not sure it is a good idea to start using this as the idea of introducing the automated code formatting was to remove subjective opinion from the process.

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/steady_state/simulation/full_function_forks/CoordinatedForkSanityTest.java
```diff
@@ -190,7 +191,7 @@ public void sanity_tests_should_pass() {
                     // verify that at epoch 21 we've successfully switched to the fork with voting
                     // (idx 3)
                     if (!verifyCurrentFork(network, forks.get(3))) {
-                      reportError.accept("Expected to be at a different fork (3) at epoch 20");
+                      reportError.accept("Expected to be at a different fork (3) at epoch 21");
```
diegomrsantos, line 194: Was it a bug in the code or the comment was wrong?

LukasGasior1: the error message was wrong

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/steady_state/simulation/full_function_forks/CoordinatedForkSanityTest.java
```diff
@@ -216,8 +217,8 @@ public void sanity_tests_should_pass() {
 
   private void updateValidatorWithLatestFork(RunningNetwork network, BFTNode node) {
     final var forks = network.getInstance(Forks.class, node);
-    final var maybeForkVoteHash =
-        forks.getCandidateFork().map(f -> ForkConfig.voteHash(node.getKey(), f));
+    final var maybeForkVote =
```
diegomrsantos, line 220: I would remove "maybe"from this variable as there's no Maybe type in Java afaik.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/model/entities/ValidatorEntity.java
```diff
@@ -111,24 +116,46 @@ public void overwriteDataObject(
       updateValidatorFee(builder, preparedValidatorFee, config);
     } else if (dataObject instanceof ValidatorMetadata metadata) {
       updateMetadata(builder, metadata);
+      updateValidatorSystemMetadataVote(builder);
     } else if (dataObject instanceof ValidatorAllowDelegation allowDelegation) {
       updateAllowDelegation(builder, allowDelegation);
     } else if (dataObject
         instanceof com.radixdlt.api.core.openapitools.model.ValidatorSystemMetadata metadata) {
-      updateValidatorSystemMetadata(builder, metadata);
+      if (metadata.getData().isEmpty()) {
```
diegomrsantos, line 124: Could we improve the readability of this code? This long if/else is quite hard to read. Maybe change it to a switch?

diegomrsantos: I meant the whole if/else block not only the new addition. IntelliJ recommends that at line 110.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/model/entities/ValidatorEntity.java
```diff
@@ -111,24 +116,46 @@ public void overwriteDataObject(
...
-      TxBuilder builder,
-      com.radixdlt.api.core.openapitools.model.ValidatorSystemMetadata metadata) {
+  private void updateValidatorSystemMetadataVote(TxBuilder builder) {
     builder.down(
         com.radixdlt.application.validators.state.ValidatorSystemMetadata.class, validatorKey);
     builder.up(
         new com.radixdlt.application.validators.state.ValidatorSystemMetadata(
-            validatorKey, Bytes.fromHexString(metadata.getData())));
+            validatorKey,
+            maybeCandidateFork
```
diegomrsantos, line 145: I would replace the "maybe" by "optional" or remove this prefix altogether.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/model/entities/ValidatorEntity.java
```diff
@@ -111,24 +116,46 @@ public void overwriteDataObject(
...
+                .map(
+                    candidateFork ->
+                        CandidateForkVote.create(validatorKey, candidateFork).payload())
+                .orElseGet(HashUtils::zero256)
+                .asBytes()));
+  }
+
+  private void updateValidatorSystemMetadataWithdrawVote(TxBuilder builder) {
+    builder.down(
+        com.radixdlt.application.validators.state.ValidatorSystemMetadata.class, validatorKey);
```
diegomrsantos, line 155: why `com.radixdlt.application.validators.state.` is necessary?

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/steady_state/simulation/full_function_forks/CoordinatedForkSanityTest.java
diegomrsantos: maybe it would be better to replace the pair of numbers by variables.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/system/api.yaml
```diff
@@ -130,10 +130,10 @@ components:
       type: object
       required:
         - network_status
-        - current_fork_hash
+        - current_fork_name
```
diegomrsantos, line 133: Still don't think this should be in the health endpoint, but in a dedicated endpoint instead.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/system/openapitools/model/HealthResponse.java
```diff
@@ -153,29 +153,29 @@ public void setNetworkStatus(NetworkStatusEnum networkStatus) {
...
   /**
-   * Get currentForkHash
-   * @return currentForkHash
+   * Get currentForkName
+   * @return currentForkName
    **/
   @javax.annotation.Nonnull
   @ApiModelProperty(required = true, value = "")
-  @JsonProperty(JSON_PROPERTY_CURRENT_FORK_HASH)
+  @JsonProperty(JSON_PROPERTY_CURRENT_FORK_NAME)
```
diegomrsantos, line 167: I see that sometimes the "hash" suffix has been replaced by "name" and sometimes it  has just been removed, like at line 239.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/CandidateForkVotesPostProcessor.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+/** Adds forksVotingResults at epoch boundary to the result metadata. */
+public final class CandidateForkVotesPostProcessor implements PostProcessor<LedgerAndBFTProof> {
+  /**
+   * Specifies a minimum stake that needs to vote for a fork to be added to the resulting stored
+   * map. Acts as a storage size optimization so that we don't store insignificant entries.
+   */
+  private static final short VOTES_THRESHOLD_TO_STORE_RESULT = 5000; /* 50.00% */
```
diegomrsantos, line 98: Maybe you could add a comment explaining why 5000 means 50%?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/CandidateForkVotesPostProcessor.java
```diff
@@ -0,0 +1,192 @@
...
+    final var validatorSet = ledgerAndBFTProof.getProof().getNextValidatorSet().orElseThrow();
+    final var totalPower = validatorSet.getTotalPower();
+    final var totalPowerVotedMap = totalStakePowerVotedByFork(engineStore, validatorSet);
+
+    return totalPowerVotedMap.entrySet().stream()
+        .map(
+            e -> {
+              final var percentagePower =
+                  e.getValue()
+                      .multiply(UInt256.from(10000))
```
diegomrsantos, line 132: It'd be good to extract this number to a descriptive variable and maybe also add a comment explaining its meaning.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/CandidateForkVotesPostProcessor.java
```diff
@@ -0,0 +1,192 @@
...
+      return metadata;
+    }
+  }
+
+  private ImmutableSet<ForkVotingResult> prepareForksVotingResults(
+      EngineStore.EngineStoreInTransaction<LedgerAndBFTProof> engineStore,
+      LedgerAndBFTProof ledgerAndBFTProof) {
+    final var nextEpoch = ledgerAndBFTProof.getProof().getEpoch() + 1;
+    final var validatorSet = ledgerAndBFTProof.getProof().getNextValidatorSet().orElseThrow();
+    final var totalPower = validatorSet.getTotalPower();
```
diegomrsantos, line 124: What is total power? Do we have a power concept in the system?

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/steady_state/simulation/full_function_forks/CoordinatedForkSanityTest.java
LukasGasior1: I prefer to make it clear when the var is optional, I suggest either `maybeXyz` or `xyzOpt`

radixdlt-core/radixdlt/src/main/java/com/radixdlt/RadixNodeModule.java
```diff
@@ -267,8 +269,20 @@ protected void configure() {
...
-      log.info("Using testing forks");
-      install(new TestingForksModule());
+      int forkConfigVersion =
+          properties.get(TESTING_FORKS_VERSION_KEY) != null
+                  && !properties.get(TESTING_FORKS_VERSION_KEY).isBlank()
+              ? properties.get(TESTING_FORKS_VERSION_KEY, 2)
+              : 2;
+      log.info("Using testing fork config, version {}", forkConfigVersion);
+      switch (forkConfigVersion) {
+        case 1 -> install(new TestingForksModuleV1());
```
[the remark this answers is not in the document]
LukasGasior1: @Theoklitos ^

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/model/entities/ValidatorEntity.java
LukasGasior1: one is an openapi dto and one the actual domain class, can't import both under and there are no aliases in java

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/CandidateForkVotesPostProcessor.java
LukasGasior1: yup, an abstraction for stake, see `BFTValidatorSet`

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/steady_state/simulation/full_function_forks/CoordinatedForkSanityTest.java
diegomrsantos: I'm not sure this is a common pattern in Java, when using an IDE we can easily hover the name or use auto complete and see this is an `Optional` type. This is something we can discuss as a team.  But in the case of using it, I believe it is better to use a name similar to the underlying type.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/model/entities/ValidatorEntity.java
diegomrsantos: Is it possible to customize the name of the generated class?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/CandidateForkVotesPostProcessor.java
diegomrsantos: are they synonyms?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/forks/CandidateForkConfig.java
```diff
@@ -64,59 +64,35 @@
...
+    implements ForkConfig {
 
-  public CandidateForkConfig(
-      String name, HashCode hash, RERules reRules, int requiredStake, long minEpoch) {
-    this.name = name;
-    this.hash = hash;
-    this.reRules = reRules;
-    this.requiredStake = requiredStake;
-    this.minEpoch = minEpoch;
+  public static record Threshold(short requiredStake, int numEpochsBeforeEnacted) {
```
diegomrsantos, line 79: should we add validation for `requiredStake`?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/forks/CandidateForkVote.java
```diff
@@ -62,61 +62,79 @@
...
-      throws PostProcessorException {
-    if (metadata.getProof().getNextValidatorSet().isPresent()) {
-      return metadata.withValidatorsSystemMetadata(
-          getValidatorsSystemMetadata(engineStore, metadata));
-    } else {
-      return metadata;
-    }
+public record CandidateForkVote(HashCode payload) {
+  public static final HashCode FORK_VOTE_NONCE =
+      HashUtils.sha256("olympia".getBytes(StandardCharsets.US_ASCII));
```
diegomrsantos, line 80: why not UTF-8?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/forks/ForkBuilder.java
```diff
@@ -64,51 +64,64 @@
...
-import com.google.common.hash.HashCode;
+import com.google.common.collect.ImmutableSet;
 import java.util.Optional;
 
 /**
  * An intermediate class used for creating fork configurations. Allows to modify various config
  * parameters before the actual fork config object is created.
  */
 public final class ForkBuilder {
+  public static record FixedEpochForkBuildersOpts(long epoch) {}
```
diegomrsantos, line 75: Could you please explain what "BuildersOpts" mean?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/forks/ForkBuilder.java
```diff
@@ -64,51 +64,64 @@
...
-      HashCode hash,
-      long epoch,
-      Optional<Integer> requiredStake,
+      Optional<FixedEpochForkBuildersOpts> fixedEpochForkBuildersOpts,
+      Optional<CandidateForkBuildersOpts> candidateForkBuildersOpts,
       RERulesVersion reRulesVersion,
       RERulesConfig reRulesConfig) {
+    if (fixedEpochForkBuildersOpts.isEmpty() == candidateForkBuildersOpts.isEmpty()) {
+      throw new IllegalArgumentException(
+          "ForkBuilder must specify either a fixed epoch or candidate opts");
```
diegomrsantos, line 119: Could you please explain what a "candidate opts" is?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/forks/Forks.java
```diff
@@ -147,9 +144,9 @@ public static Forks create(Set<ForkConfig> forks) {
     return new Forks(fixedEpochForks, maybeCandidateFork);
   }
 
-  private static boolean ensureUniqueHashes(Set<ForkConfig> forks) {
-    final var hashesSet = forks.stream().map(ForkConfig::hash).collect(Collectors.toSet());
-    return forks.size() == hashesSet.size();
+  private static boolean ensureUniqueNames(Set<ForkConfig> forks) {
+    final var namesSet = forks.stream().map(ForkConfig::name).collect(Collectors.toSet());
```
diegomrsantos, line 148: Still in the topic of the Hungarian Notation, I believe it is mostly considered a bad practice https://en.wikipedia.org/wiki/Hungarian_notation#Notable_opinions.

LukasGasior1: I think it's justified in this particular case, using Set specifically is the mechanism that actually checks for duplicates. Calling it simply `names` would provide much less context imo

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/steady_state/simulation/full_function_forks/CoordinatedForkSanityTest.java
LukasGasior1: I'm not sure, there are 5 different epochs where the test does things, I'd need to add vars for all of them

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/steady_state/simulation/full_function_forks/CoordinatedForkSanityTest.java
LukasGasior1: yup, let's discuss

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/model/entities/ValidatorEntity.java
LukasGasior1: done (changed to swtich with pattern matching)

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/model/entities/ValidatorEntity.java
LukasGasior1: I don't know, but even if so, I'd rather keep the name as is to match the api specs

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/system/openapitools/model/HealthResponse.java
LukasGasior1: that's because in some cases it's just a name (string) and in other cases it's an object with a name and other fields

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/CandidateForkVotesPostProcessor.java
LukasGasior1: added a comment in `CandidateForkConfig.Threshold` and a pointer here

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/CandidateForkVotesPostProcessor.java
LukasGasior1: To some extent but they're on different abstraction levels. The consensus layer doesn't really know that power is backed by stake

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/forks/CandidateForkConfig.java
LukasGasior1: good point, done

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/forks/CandidateForkVote.java
LukasGasior1: it doesn't really matter, nonce is just some constant seq of bytes

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/forks/ForkBuilder.java
LukasGasior1: different options for the two types of fork configs (candidate and fixed epoch)

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/model/entities/ValidatorEntity.java
diegomrsantos: I see. In this case, we are already importing `com.radixdlt.application.validators.state.ValidatorSystemMetadata`, we can just use `ValidatorSystemMetadata` for this class and `com.radixdlt.api.core.openapitools.model.ValidatorSystemMetadata` for the other one.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/system/openapitools/model/HealthResponse.java
diegomrsantos: Does that mean that `HealthResponseUnknownReportedForksHashes` shouldn't be named like that before and the fix was just to remove the suffix?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/CandidateForkVotesPostProcessor.java
diegomrsantos: Couldn't we use a percentage value between 0.01 and 100.00 and parse it as a `BigDecimal`?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/CandidateForkVotesPostProcessor.java
diegomrsantos: I was wondering if we could just use "stake" in both cases.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/forks/Forks.java
diegomrsantos: But what the real value of doing that in a statically typed language? If a developer wants to know the type of a variable it can be easily found.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineModule.java
```diff
@@ -93,13 +93,13 @@ private CurrentForkView currentForkView(
       CommittedReader committedReader, ForksEpochStore forksEpochStore, Forks forks) {
     forks.init(committedReader, forksEpochStore);
 
-    final var maybeLatestForkHash =
-        forksEpochStore.getEpochsForkHashes().entrySet().stream()
+    final var maybeLatestForkName =
```
diegomrsantos, line 96: Here we are still using "maybe".

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
```diff
@@ -395,27 +395,7 @@ public StateComputerResult prepare(
...
-              this.radixEngine.replaceConstraintMachine(
-                  rules.getConstraintMachineConfig(),
-                  rules.getSerialization(),
-                  rules.getActionConstructors(),
-                  rules.getPostProcessor(),
-                  rules.getParser());
-              this.epochCeilingView = rules.getMaxRounds();
-              this.maxSigsPerRound = rules.getMaxSigsPerRound();
-            });
+    result.getMetadata().getNextForkName().ifPresent(this::forkRadixEngine);
```
diegomrsantos, line 398: I don't really have a suggestion, but it seems a bit weird this is mutating state. What do you think?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/berkeley/BerkeleyLedgerEntryStore.java
```diff
@@ -416,60 +394,52 @@ private void storeMetadata(Transaction dbTxn, LedgerAndBFTProof ledgerAndBFTProo
...
-  private void storeEpochValidatorsSystemMetadata(
-      Transaction dbTxn, long epoch, ImmutableList<RawSubstateBytes> validatorSystemMetadata) {
-    removeOldEpochValidatorsSystemMetadataEntries(dbTxn, epoch);
-    final var key = new DatabaseEntry(Longs.toByteArray(epoch));
-    validatorSystemMetadata.forEach(
-        data ->
-            validatorsSystemMetadataDatabase.put(dbTxn, key, new DatabaseEntry(data.getData())));
+  private void storeForksVotingResults(
+      Transaction dbTxn, ImmutableSet<ForkVotingResult> forksVotingResults) {
+    forksVotingResults.forEach(
```
diegomrsantos, line 407: In cases like that, a simple for-each loop is more readable imho.

build.gradle
LukasGasior1: I can remove it if you insist, although I'm pretty sure it makes that test code objectively harder to read

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/core/model/entities/ValidatorEntity.java
LukasGasior1: done

radixdlt-core/radixdlt/src/main/java/com/radixdlt/api/system/openapitools/model/HealthResponse.java
LukasGasior1: yeah, it should be `HealthResponseUnknownReportedForks` from the beginning

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/CandidateForkVotesPostProcessor.java
LukasGasior1: why?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/forks/ForkBuilder.java
LukasGasior1: the builder is the outer class, this is just a container for two different types of fork parameters

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/forks/Forks.java
LukasGasior1: the real value is that it explicitly states that the variable is a set, and it must remain a set because this is a part of the implementation, if the name was just `names` it wouldn't be obvious that it can't just be replaced with a list

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
LukasGasior1: what do you mean? the state is usually supposed to be modified ;)

radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/berkeley/BerkeleyLedgerEntryStore.java
LukasGasior1: I think it's a matter of taste, but in general we try to follow the functional style where possible

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/steady_state/simulation/full_function_forks/CoordinatedForkSanityTest.java
LukasGasior1: The goal here isn't really to provide the information about the type, but rather to capture what the value really is.
If the names was just `forkVote` it would mean that the value represents a fork vote, which is not always true. The value can represent either a fork vote or lack of thereof. The type doesn't matter, it could be any container (Optional, Maybe, etc) or an ADT (like an enum [if Java had something like that] ` { ForkVote(HashCode value), NoVote }`) or it could be a plain nullable value - in all these cases I'd still like to capture this information in the name of the variable (with either a `maybe` prefix or `Opt` suffix).

LukasGasior1: also, it removes the name conflict if at some later point you want to extract it to a non-nullable value:
```
val xyzOpt = ...
val xyz = xyzOpt.orElse(defaultXyz);
```

build.gradle
LukasGasior1: removed the flag

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/steady_state/simulation/full_function_forks/CoordinatedForkSanityTest.java
diegomrsantos: That's exactly what an `Optional<CandidateForkVote>` conveys, either a fork vote or nothing.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/RadixNodeModule.java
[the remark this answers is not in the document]
Theoklitos: will do

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/CandidateForkVotesPostProcessor.java
diegomrsantos: Cause stake is a well defined concept in the DLT domain.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/store/berkeley/BerkeleyLedgerEntryStore.java
diegomrsantos: I would suggest to be more flexible and use functional where it is more appropriate and not use it when it is not. I've seen @talekhinezh making similar suggestions before.

LukasGasior1: why do you think it's not appropriate here?

diegomrsantos: My comment was more related to the approach of trying to follow the functional style where possible. But in this case a simple for-each would be better imo. Streams seem more suitable for data processing pipelines.

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/steady_state/simulation/full_function_forks/CoordinatedForkSanityTest.java
LukasGasior1: following this reasoning, can I just call it `potato` since the type already conveys the meaning, why duplicate the same thing?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/forks/Forks.java
LukasGasior1: I've changed it to a more descriptive `namesWithoutDuplicates` that isn't tied to a specific type

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/steady_state/simulation/full_function_forks/CoordinatedForkSanityTest.java
diegomrsantos: I think it seems weird to me cause I've never seen anyone using "possibleValue" to name a variable which represents the result of getting a value from a map in Java, which can return null, for example. Seems this need only arises when using an `Optional` type which is not that different conceptually, only more explicit and safer.

LukasGasior1: I think not naming things correctly is one of the reasons why null reference is considered a billion-dollar mistake ;) Imagine how much less likely it would be to miss a null check if the name clearly stated that the value might or might not be there. I'd definitely go for `maybeXyz` (or `xyzOpt`) even for nullable values, but luckily we have Optional.

PR #646, opened 2022-02-28
Preparation for NT258, refactoring and cleanup of TxActions and REEvent

siy:
No changes in functionality, only refactoring changes.


radixdlt-engine/src/main/java/com/radixdlt/atom/TxAction.java
```diff
@@ -64,5 +64,216 @@
...
+      String name,
+      String description,
+      String iconUrl,
+      String tokenUrl,
+      UInt256 supply)
+      implements TxAction {
+    public CreateFixedToken {
+      ensureResourceAddressIsHashedKey(resourceAddr);
+      ensureNonZeroSupply(supply);
+      requireAllNonNull(accountAddr, symbol, name, description, iconUrl, tokenUrl);
```
dhedey, line 99: I'd prefer to have separate lines for `requireNonNull`

radixdlt-engine/src/main/java/com/radixdlt/atom/TxAction.java
```diff
@@ -64,5 +64,216 @@
...
+        String name,
+        String description,
+        String iconUrl,
+        String tokenUrl,
+        ECPublicKey owner) {
+      this.resourceAddress = resourceAddress;
+      this.symbol = symbol.toLowerCase();
+      this.owner = owner;
+
+      this.name = asEmptyIfNull(name);
```
dhedey, line 124: I notice we validate this differently from the fixed token (because we're trying not to change behaviour) - do you think it might be worth explaining why or investigating whether we can safely do null checks on this instead of replacing with empty?

radixdlt-engine/src/main/java/com/radixdlt/atom/TxAction.java
```diff
@@ -64,5 +64,216 @@
...
+
+  record FeeReserveComplete(REAddr toAddr) implements TxAction {
+    public FeeReserveComplete {
+      ensureAddressIsAccount(toAddr);
+    }
+  }
+
+  record FeeReservePut(REAddr fromAddr, UInt256 amount) implements TxAction {}
+
+  record FeeReserveTake(REAddr toAddr, UInt256 amount) implements TxAction {}
```
dhedey, line 141: Similarly, do you think we should assert these are account addresses here (similar to FeeReserveComplete?) What impacts would that have?

radixdlt-engine/src/main/java/com/radixdlt/atom/TxAction.java
```diff
@@ -64,5 +64,216 @@
...
+    public MintToken {
+      ensureNonZeroAmount(amount);
+    }
+  }
+
+  record NextEpoch(long timestamp) implements TxAction {}
+
+  record NextRound(
+      long view, boolean isTimeout, long timestamp, LongFunction<ECPublicKey> leaderMapping)
+      implements TxAction {}
```
dhedey, line 153: I think we're missing the nice ToString on this Action - although I'm not sure where it's used

radixdlt-engine/src/main/java/com/radixdlt/atom/TxAction.java
```diff
@@ -64,5 +64,216 @@
...
+import com.radixdlt.crypto.ECPublicKey;
+import com.radixdlt.identifiers.REAddr;
+import com.radixdlt.utils.Bytes;
+import com.radixdlt.utils.UInt256;
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.function.LongFunction;
+
+/** RadixEngine actions. */
+public sealed interface TxAction {
```
dhedey, line 79: Quite a few of the actions under here are missing their custom ToString methods

dhedey (approved):
Looks good

radixdlt-engine/src/main/java/com/radixdlt/atom/TxnConstructionRequest.java
```diff
@@ -130,11 +124,6 @@ public TxnConstructionRequest actions(List<TxAction> actions) {
     return this;
   }
 
-  public TxnConstructionRequest createMutableToken(MutableTokenDefinition def) {
```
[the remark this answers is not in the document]
siy: yes

radixdlt-engine/src/main/java/com/radixdlt/atom/TxAction.java
siy: I guess this somehow reflects business requirements, but I'm not aware about any relevant documents.

radixdlt-engine/src/main/java/com/radixdlt/atom/TxAction.java
siy: For each `record` compiler generates `toString()` method which is at least as good as one we had before. For example, here is output for `CreateSystem` action:
```
CreateSystem[timestamp=1]
```

radixdlt-engine/src/main/java/com/radixdlt/atom/TxAction.java
siy: See comment above.

radixdlt-engine/src/main/java/com/radixdlt/atom/TxAction.java
siy: Done

radixdlt-engine/src/main/java/com/radixdlt/atom/TxAction.java
siy: I think this is a separate issue. The goal of this PR is to clean up code while preserving existing functionality.
I've added a dedicated ticket to make all validations consistent: https://radixdlt.atlassian.net/browse/NT-276

PR #649, opened 2022-03-07
Implement update address book on successful connection

siy:


michalzajda (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/addressbook/AddressBook.java
```diff
@@ -229,10 +225,44 @@ private boolean sameNetworkHrp(RadixNodeUri uri) {
   }
 
   public void addOrUpdatePeerWithSuccessfulConnection(RadixNodeUri radixNodeUri) {
+    this.cleanupPeerEntriesFromRemoteHostPort(radixNodeUri.getHost(), radixNodeUri.getPort());
```
LukasGasior1, line 228: I think we might only want to do this if `peerChannel.isInitiator == true`, otherwise the port is just a randomly assigned one

siy: Fixed in https://github.com/radixdlt/radixdlt/pull/652

PR #655, opened 2022-03-15
Forward origin peer to MempoolAddSuccess event

LukasGasior1:
Forward origin peer to MempoolAddSuccess event so that the tx is not relayed back

### DIFF
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
@@ -226,7 +226,7 @@ public final class RadixEngineStateComputer implements StateComputer {
         .forEach(
             txn -> {
               try {
-                addToMempool(txn);
+                addToMempool(txn, origin);
               } catch (MempoolRejectedException ignored) {
               }
             });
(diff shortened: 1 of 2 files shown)


radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
```diff
@@ -226,7 +226,7 @@ public void addToMempool(MempoolAdd mempoolAdd, @Nullable BFTNode origin) {
         .forEach(
             txn -> {
               try {
-                addToMempool(txn);
+                addToMempool(txn, origin);
               } catch (MempoolRejectedException ignored) {
               }
```
siy, line 231: I understand that the exception here is ignored, but perhaps we can add at least logging here (perhaps even at debug level)? I feel uncomfortable if something might happen without at least leaving any signs in logs.

siy (approved):
LGTM with minor comment

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
LukasGasior1: sure

LukasGasior1: done

PR #663, opened 2022-03-31
Fix deadlock related to how the PendingChannelsManager completes the futures

LukasGasior1:
The fix is to release the lock before calling `channelFuture.completeExceptionally`

### DIFF
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PendingOutboundChannelsManager.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PendingOutboundChannelsManager.java
@@ -107,6 +107,8 @@ public final class PendingOutboundChannelsManager {
     var existingChannelFutureOpt =
         Optional.ofNullable(this.pendingChannels.putIfAbsent(remoteNodeId, newChannelFuture));
 
+    // There was no existing channel future that has already been initialized for this node ID
+    //   so we init a new outbound connection
     if (existingChannelFutureOpt.isEmpty()) {
       this.peerOutboundBootstrap.initOutboundConnection(uri);
       this.timeoutEventDispatcher.dispatch(
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PendingOutboundChannelsManager.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PendingOutboundChannelsManager.java
@@ -73,13 +73,11 @@ import com.radixdlt.network.p2p.PeerEvent.PeerHandshakeFailed;
 import com.radixdlt.network.p2p.transport.PeerChannel;
 import com.radixdlt.network.p2p.transport.PeerOutboundBootstrap;
 import java.io.IOException;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 import javax.inject.Inject;
 
 public final class PendingOutboundChannelsManager {
@@ -87,8 +85,8 @@ public final class PendingOutboundChannelsManager {
   private final PeerOutboundBootstrap peerOutboundBootstrap;
   private final ScheduledEventDispatcher<PeerOutboundConnectionTimeout> timeoutEventDispatcher;
   private final EventDispatcher<PeerEvent> peerEventDispatcher;
-  private final Map<NodeId, CompletableFuture<PeerChannel>> pendingChannels = new ConcurrentHashMap<>() {
-  };
+  private final Map<NodeId, CompletableFuture<PeerChannel>> pendingChannels =
+      new ConcurrentHashMap<>() {};
 
   @Inject
   public PendingOutboundChannelsManager(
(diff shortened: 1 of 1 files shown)


radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PendingOutboundChannelsManager.java
```diff
@@ -128,37 +128,43 @@ public PendingOutboundChannelsManager(
   }
 
   private void handlePeerConnected(PeerConnected peerConnected) {
+    final var channel = peerConnected.channel();
+    final Optional<CompletableFuture<PeerChannel>> channelFutureOpt;
     synchronized (lock) {
```
diegomrsantos, line 133: Why do we need a lock here? Can't we use a ConcurrentMap?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PendingOutboundChannelsManager.java
```diff
@@ -128,37 +128,43 @@ public PendingOutboundChannelsManager(
...
-      final var channel = peerConnected.channel();
-      final var maybeFuture = this.pendingChannels.remove(channel.getRemoteNodeId());
-      if (maybeFuture != null) {
-        maybeFuture.complete(channel);
-      }
+      channelFutureOpt =
```
diegomrsantos, line 134: Why this fixes the deadlock? Is there another code running in this same thread preventing the Future of being completed?

dhedey (approved):
LGTM 👍

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PendingOutboundChannelsManager.java
dhedey: Possibly - sounds like a reasonable idea, it depends what the lock is protecting - is it just the thread safety of the map, or is the lock protecting other logical guarantees (eg about the number of connection attempts to a single node in connectTo).

On that note - @LukasGasior1  the lock in `connectTo` feels overly locked, and calls a `dispatch` inside the lock, which I don't like the look of, although dispatch is possibly thread-safe?  But I don't fully understand the implications of the locking here and what it's supposed to prevent, but if it can be reduced to just locking the Map, and so be replaced by a ConcurrentMap, that's not a bad idea :)

LukasGasior1: @diegomrsantos we could use a ConcurrentMap and its `computeIfAbsent` method in `connectTo`, but as @dhedey pointed out, doing it with a manually managed lock allows us to extract `peerOutboundBootstrap.initOutboundConnection` and `timeoutEventDispatcher.dispatch` out of the lock in `connectTo` - which I did in the latest commit. With a ConcurrentMap these two calls would need to be in the `computeIfAbsent` callback, holding ConcurrentMap's lock.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PendingOutboundChannelsManager.java
LukasGasior1: Future's `onComplete` callbacks happen on the thread that calls the `.complete` method (which still holds the lock). The fix is to release the lock before calling the `.complete` method. See a more detailed write up in the Jira ticket.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/PendingOutboundChannelsManager.java
diegomrsantos: Not sure if I follow. It seems the only thing this lock is preventing is two concurrent removals from the map. In a concurrent map I would expect one to succeed and the other to return null and then nothing would be done.

diegomrsantos: Just to clarify, I wasn't talking about the connectTo method, but the specific one where I added the comment. I was also talking about the use of the lock in this method not the lock in general. The other methods follow a similar pattern where the synchronize block is used only to protect concurrent removals from the map.

LukasGasior1: In `connectTo` it synchronized the check whether the key is in the map, and the put operation, followed by some actions (in case the element was missing).
While the specific place where you added the comment could benefit from the ConcurrentMap, the `connectTo` method would still require an external lock (see my comment above) and I think it's better to stick to one approach rather than mixing ConcurrentMap's internal lock with an external object lock in a single class.

diegomrsantos: The deadlock was caused by a lock that could be avoided and it seems the same can be said for all the other methods. What are the disadvantages of using the `ConcurrentMap` and manually locking only where there's no trivial alternative?

diegomrsantos: Actually, reading it one more time, I can't see any difference between the `synchronised` block  in `connectTo` and using [putIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentMap.html#putIfAbsent-K-V-).

LukasGasior1: Given my comment above, how would you avoid a lock for all other method? I don't quite see it. Would you mind submitting a commit for your suggested changes?
The disadvantage to me is that, because we still need to use the synchronized block elsewhere, there's no point of using a synchronized collection. In my opinion it's easier to reason about the code if only a single locking mechanism is used.
Additionally, if we used a concurrent collection, there would be some performance impact in `connectTo`, because of duplicated synchronization (internal in the collection and the outer synchronized block) [the impact is probably quite low, but in any case, it's one more argument on my side :laughing: ].

LukasGasior1: Currently `peerOutboundBootstrap.initOutboundConnection` and `timeoutEventDispatcher.dispatch` are called after releasing the lock. With `putIfAbsent` we'd need to call these in the absent callback, while still holding the lock.
And we can't set a non-final flag variable from within the callback.

LukasGasior1: hmh, although this could bo probably solved by using some mutable container, for example AtomicBoolean

diegomrsantos: I submitted a commit. `putIfAbsent` is atomic. Please review if it makes sense. Is there a test that could be run?

LukasGasior1: Thanks! I think it looks good, didn't realize that `putIfAbsent` returns the previous value **or null**.

diegomrsantos (approved):
LGTM

PR #664, opened 2022-04-05
Add unit test for PendingOutboundChannelsManager thread safety

LukasGasior1:


radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/PendingOutboundChannelsManagerTest.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicReference;
+import org.json.JSONObject;
+import org.junit.Test;
+
+public final class PendingOutboundChannelsManagerTest {
+  @Test
+  public void
+      pending_outbound_channels_manager_should_return_the_same_future_when_called_from_diff_threads()
```
dhedey, line 90: Can you add a comment or java doc explaining how this test arose, and why it's important that the same future is returned from different threads?

I thought the issue was a combination of:
* Resolving a future during a lock
* That the future's callback was being executed on the wrong thread? (the P2PManagerThread instead of the Outbound message processing thread?)

I'm not sure it's obvious to me how this test shows that either issue is fixed? A comment might help :)

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/PendingOutboundChannelsManagerTest.java
```diff
@@ -0,0 +1,131 @@
...
+                new RuntimeProperties(new JSONObject(), new String[] {})),
+            mock(PeerOutboundBootstrap.class),
+            rmock(ScheduledEventDispatcher.class),
+            rmock(EventDispatcher.class));
+
+    final var remoteNodeId = NodeId.fromPublicKey(ECKeyPair.generateNew().getPublicKey());
+    final var remoteNodeUri =
+        RadixNodeUri.fromPubKeyAndAddress(1, remoteNodeId.getPublicKey(), "127.0.0.1", 30000);
+
+    for (int i = 0; i < 200; i++) {
```
diegomrsantos, line 106: Maybe we can add a comment explaining why 200

radixdlt-core/radixdlt/src/test/java/com/radixdlt/network/p2p/PendingOutboundChannelsManagerTest.java
dhedey: The test as written doesn't obviously prove to me the code as a whole can't deadlock - maybe this implication is indirect?

I think it'd be useful to capture an explanation of the implication / indirection into the codebase, so someone reviewing it in the future can understand why this test is important when they consider whether to refactor/fix/delete it.

LukasGasior1: ah, you've already answered @diegomrsantos , missed that :)

diegomrsantos: As we aren't directly using locks, probably, we don't need to care about deadlocks anymore. The critical issue is the race condition that I believe the previous lock was trying to solve. The test tries to show it doesn't happen when replacing the lock by a `ConcurrentHashMap`. It could be made more explicit with comments.

LukasGasior1: added a comment

diegomrsantos (approved):

PR #669, opened 2022-04-19
Use dedicated hosts for the integration tests

muzuke:
The PR configures the integration jobs to use a dedicated runner.


muzuke:
@diegomrsantos Seems like there isn't a lot of benefit from the runners. Should we tweak some build parameters to leverage more cores?

shambupujar (approved):

PR #670, opened 2022-04-19
Revised log levels for noisy log messages

dhedey:
PR to merge into main; Will create a separate PR to merge into the release branch.

### DIFF
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/ExceptionHandler.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/ExceptionHandler.java
@@ -109,8 +109,12 @@ public class ExceptionHandler extends ChannelDuplexHandler {
   }
 
   private void logAndCloseChannel(ChannelHandlerContext ctx, Throwable cause) {
-    var affectedEntity = mainHandler.map(Object::toString).orElse("channel");
-    log.warn("Exception on {} {}, closing channel to prevent resource leak", affectedEntity, cause);
+    if (log.isTraceEnabled()) {
+      var affectedEntity = mainHandler.map(Object::toString).orElse("(to a peer via proxy)");
+      log.trace("Closing channel {} due to {}", affectedEntity, cause);
+    }
+
+    // The channel is closed to prevent a resource leak
     ctx.close();
   }
 }
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
@@ -227,8 +227,10 @@ public final class RadixEngineStateComputer implements StateComputer {
             txn -> {
               try {
                 addToMempool(txn, origin);
+              } catch (MempoolDuplicateException ex) {
+                log.trace("Transaction {} was not added as it was already in the mempool", txn.getId());
               } catch (MempoolRejectedException ex) {
-                log.debug("Txn rejected from the mempool", ex);
+                log.debug("Transaction {} was not added to the mempool", txn.getId(), ex);
               }
             });
   }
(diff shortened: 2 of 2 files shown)


LukasGasior1 (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/ExceptionHandler.java
```diff
@@ -109,8 +109,12 @@ private void handleFailure(Future<?> future, ChannelHandlerContext ctx) {
   }
 
   private void logAndCloseChannel(ChannelHandlerContext ctx, Throwable cause) {
-    var affectedEntity = mainHandler.map(Object::toString).orElse("channel");
-    log.warn("Exception on {} {}, closing channel to prevent resource leak", affectedEntity, cause);
+    if (log.isTraceEnabled()) {
+      var affectedEntity = mainHandler.map(Object::toString).orElse("(to a peer via proxy)");
+      log.trace("Closing channel {} due to {}", affectedEntity, cause);
```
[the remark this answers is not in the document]
diegomrsantos: This one too https://stackoverflow.com/questions/7839565/logging-levels-logback-rule-of-thumb-to-assign-log-levels?noredirect=1&lq=1

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
```diff
@@ -227,8 +227,10 @@ public void addToMempool(MempoolAdd mempoolAdd, @Nullable BFTNode origin) {
             txn -> {
               try {
                 addToMempool(txn, origin);
+              } catch (MempoolDuplicateException ex) {
+                log.trace("Transaction {} was not added as it was already in the mempool", txn.getId());
```
diegomrsantos, line 231: don't we need the cause here?

dhedey: The cause just says that a duplicate transaction was there, which causes loads of useless repetition and stack traces in logs. We have all the information we need here.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/network/p2p/transport/ExceptionHandler.java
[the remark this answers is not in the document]
dhedey: It's trace because the rest of the P2P messages (which aren't suggestive of bugs or misconfigurations) are trace.

There's an argument some of these messages should be DEBUG (or, most strictly, we should have a review of our peer to peer logging and what we log. I'd suggest EG we should actually log clearly at DEBUG when a connection is opened or closed to a peer at a give address), but this isn't part of this ticket I feel.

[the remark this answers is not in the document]
dhedey: I'll raise a separate ticket to revise our default logging level, logging configuration and logging strategy more generally.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
diegomrsantos: So maybe we should log the original message or change the original message to the one created now?

dhedey: For me, there's an abstraction difference.

The exception is thrown in a place where the mempool add could have come via different paths (eg from a full node adding it themselves due to a submission, or due to a validator adding a vote transaction) - these may want to handle the exception differently, and so the message should be quite general, and describe the problem. EG if someone submits a transaction from the Core API, we'd want to report it's a duplicate; but if it comes from Node Gossip (as in this case),  that's totally expected.

In this particular catch, we know that the message is from Node Gossip, and so we're fine to discard duplicates. The message should reflect that we've discarded a duplicate and we're not worried.

PR #674, opened 2022-04-25
Add a configuration option and the implement the code which enables the file creation/update

diegomrsantos:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
```diff
@@ -184,19 +212,59 @@ public void process(
...
+    this.substateAccumulatorHashDatabase.put(
+        dbTxn, key, new DatabaseEntry(this.currentSubstateAccumulatorHash));
+  }
+
+  private void logEpochHash() {
+    logger.info(
+        "Epoch Hash: {} for epoch {}. Time spent since last epoch: {} s.",
+        Bytes.toHexString(this.currentSubstateAccumulatorHash),
+        lastEpochInDbOpt,
+        watch.elapsed().toSeconds());
```
dhedey, line 226: If the node turns on mid transaction, this watch isn't so accurate. But maybe that's okay? What's the intended use case for this watch?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
```diff
@@ -153,29 +187,23 @@ public void process(
     for (REStateUpdate reStateUpdate : txn.stateUpdates().toList()) {
       subStateBytes = Arrays.concatenate(subStateBytes, getBytes(reStateUpdate));
       if (reStateUpdate.getParsed() instanceof EpochData epochData) {
-        epoch = epochData.epoch();
+        lastEpochInDbOpt = Optional.of(epochData.epoch());
```
dhedey, line 190: Out of interest - why is this a class variable and not a method variable?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
```diff
@@ -184,19 +212,59 @@ public void process(
...
+
+  private OperationStatus persistEpochHash(Transaction dbTxn) {
+    return epochHashDatabase.put(
+        dbTxn,
+        new DatabaseEntry(Longs.toByteArray(lastEpochInDbOpt.orElseThrow())),
+        new DatabaseEntry(this.currentSubstateAccumulatorHash));
+  }
+
+  private void handleEpochHashAccumulatorFileUpdate() {
+    // we do not overwrite hashes and only append in a sequential way
```
dhedey, line 237: I believe the team consensus is that comments should start with a capital letter.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
```diff
@@ -215,6 +283,55 @@ private byte[] getBytes(Particle particle) {
...
+  }
+
+  private void openEpochsHashFile() {
+    var projectFolder = System.getProperty("user.dir");
+    File file = new File(projectFolder, EPOCH_HASH_FILE_PATH);
+    try {
+      this.lastEpochInFileOpt = getLastEpochInFile(file);
+      assertEpochInTheDbIsNotGreaterThanEpochInTheFile(lastEpochInFileOpt);
+      CharSink charSink = Files.asCharSink(file, StandardCharsets.UTF_8, FileWriteMode.APPEND);
+      writer = charSink.openStream();
```
dhedey, line 332: Can we give this a better name than writer - eg epochsHashFileWriter ?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
```diff
@@ -184,19 +212,59 @@ public void process(
...
+        dbTxn,
+        new DatabaseEntry(Longs.toByteArray(lastEpochInDbOpt.orElseThrow())),
+        new DatabaseEntry(this.currentSubstateAccumulatorHash));
+  }
+
+  private void handleEpochHashAccumulatorFileUpdate() {
+    // we do not overwrite hashes and only append in a sequential way
+    if (lastEpochInFileOpt.orElse(0L) == lastEpochInDbOpt.orElse(0L) - 1L) {
+      try {
+        writer.write(
```
dhedey, line 240: Can we flush the writer to disk before continuing? Else we might corrupt things

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
```diff
@@ -215,6 +283,55 @@ private byte[] getBytes(Particle particle) {
...
+      }
+    } else if (this.lastEpochInDbOpt.map(it -> it > lastEpochInFileOpt.get()).orElse(false)) {
+      // we have an epoch in the file and need to check it is not greater than the one in the db (if
+      // any)
+      isValid = false;
+    }
+    if (!isValid) {
+      throw new IllegalStateException(
+          String.format(
+              "Epoch in db %s is greater than epoch in file %s",
```
dhedey, line 307: If someone's node crashes with this exception, what should they do? Can you explain in the error message the likely cause/fix?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
```diff
@@ -67,77 +67,111 @@
...
+      "radixdlt-core/radixdlt/src/main/resources/epoch-hash";
+  public static final String EPOCH_HASH_FILE_SEPARATOR = "=";
 
   private Database substateAccumulatorHashDatabase;
   private Database epochHashDatabase;
 
   private byte[] currentSubstateAccumulatorHash;
-  private long epoch;
+  private Optional<Long> lastEpochInDbOpt;
+  private Optional<Long> lastEpochInFileOpt;
```
dhedey, line 108: I'm not sure if we need to save these against the class, or if they can be separate?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
```diff
@@ -67,77 +67,111 @@
...
+  private static final byte[] SUBSTATE_ACCUMULATOR_HASH_KEY =
       "substate_accumulator_hash_key".getBytes(StandardCharsets.UTF_8);
+  private static final String EPOCH_HASH_FILE_PATH =
+      "radixdlt-core/radixdlt/src/main/resources/epoch-hash";
+  public static final String EPOCH_HASH_FILE_SEPARATOR = "=";
 
   private Database substateAccumulatorHashDatabase;
   private Database epochHashDatabase;
 
   private byte[] currentSubstateAccumulatorHash;
```
dhedey, line 106: If an exception happens half way through ingesting a batch; this hash could be out of sync with the database, causing incorrect hashes to be written - can we instead just read from the database when we need to instead of having this cached on the class?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
```diff
@@ -67,77 +67,111 @@
...
-                null,
-                "radix.epoch_hash",
-                new DatabaseConfig()
-                    .setAllowCreate(true)
-                    .setTransactional(true)
-                    .setKeyPrefixing(true)
-                    .setBtreeComparator(lexicographicalComparator()));
+    this.substateAccumulatorHashDatabase = openSubstateAccumulatorHashDatabase(dbEnv);
+    this.epochHashDatabase = openEpochHashDatabase(dbEnv);
     this.currentSubstateAccumulatorHash =
```
dhedey, line 128: I think we're still missing the check we discussed yesterday - that we only allow the accumulator to start if it is actually the genesis transaction.

Also, it's more consistent with the other accumulators we use to start with a hash of 32 zeros - byte[32], not byte[0].

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
```diff
@@ -67,77 +67,111 @@
...
 import org.bouncycastle.util.Arrays;
+import org.bouncycastle.util.encoders.Hex;
 
 public class BerkeleySubstateAccumulatorHashStore implements BerkeleyAdditionalStore {
 
   private static final Logger logger = LogManager.getLogger();
-  protected static final byte[] SUBSTATE_ACCUMULATOR_HASH_KEY =
+  private static final byte[] SUBSTATE_ACCUMULATOR_HASH_KEY =
       "substate_accumulator_hash_key".getBytes(StandardCharsets.UTF_8);
+  private static final String EPOCH_HASH_FILE_PATH =
```
dhedey, line 99: This is going to need to depend on the environment (different environments will have different epoch hashes), so maybe we can configure it as a configuration value/environment variable?

dhedey (changes requested):
Looks good, a few areas to consider - particularly around atomicity if there are failures midway through transaction batch ingestion.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
diegomrsantos: What do you mean by mid transaction? I believe the sentence "Time spent since last epoch" won't be accurate when the node starts in the middle of an Epoch, but I think it is not a big issue. The intent is to measure the time execution overhead introduced by this feature.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
```diff
@@ -153,29 +187,23 @@ public void process(
...
-          new DatabaseEntry(Longs.toByteArray(epoch)),
-          new DatabaseEntry(this.currentSubstateAccumulatorHash));
+      persistEpochHash(dbTxn);
       if (logger.isInfoEnabled()) {
-        logger.info(
-            "Epoch Hash: {} for epoch {}. Time spent since last epoch: {} s.",
-            Bytes.toHexString(this.currentSubstateAccumulatorHash),
-            epoch,
-            watch.elapsed().toSeconds());
+        logEpochHash();
```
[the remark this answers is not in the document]
dhedey: If I come along and think "oh, I want to change the log level of this message to a DEBUG", I would search the message string, find the `logEpochHash` method and (eg) change it to a `logger.debug` - and would have missed the fact that I also need to change the `logger.isInfoEnabled` line. They are intrinsically tied together, part of the same logging abstraction (the logger.isInfoEnabled is just a performance optimisation), so should be kept together in my mind.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
dhedey: Ahh, so it just tracks the time spent doing this substate stuff? Cool - in which case, perhaps it should be renamed to something like (eg) `timeSpentOnSubstateAccumulatorThisEpoch` instead of `watch`?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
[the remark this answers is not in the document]
diegomrsantos: That's a good reason. Should I move the `logger.isInfoEnabled` to `logEpochHash` to keep the `process` cleaner?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
diegomrsantos: Agree, gonna change it.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
diegomrsantos: Could you elaborate more on this?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
diegomrsantos: Could you please elaborate more on this?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
diegomrsantos: I believe the only fix is to clean the ledger. Is it good "Please delete the content in the db folder and try again."?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
diegomrsantos: The check is being done here https://radixdlt.atlassian.net/browse/NT-337.

I'll change the initial hash.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
```diff
@@ -153,29 +187,23 @@ public void process(
     for (REStateUpdate reStateUpdate : txn.stateUpdates().toList()) {
       subStateBytes = Arrays.concatenate(subStateBytes, getBytes(reStateUpdate));
```
diegomrsantos, line 188: @dhedey should I use a 32 zero byte array as the `subStateBytes` initial value too?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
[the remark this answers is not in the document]
dhedey: Sure, sounds good 👍

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
diegomrsantos: Calling the Java's flush method doesn't avoid this issue, but it is not a big deal in this case as we agreed. I'll add the flush.

dhedey (approved):

PR #681, opened 2022-05-02
Epoch hash verification

diegomrsantos:
Verify the computed epoch hash against the ones in the file

### DIFF
--- a/radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
+++ b/radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
@@ -97,18 +97,30 @@ public class BerkeleySubstateAccumulatorHashStore implements BerkeleyAdditionalS
 
   private static final Logger logger = LogManager.getLogger();
 
-  public static final String UPDATE_EPOCH_HASH__FILE_ENABLE_PROPERTY_NAME =
+  public static final String UPDATE_EPOCH_HASH_FILE_ENABLE_PROPERTY_NAME =
       "update_epoch_hash_file.enable";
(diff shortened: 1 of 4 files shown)


radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
```diff
@@ -95,66 +97,101 @@ public class BerkeleySubstateAccumulatorHashStore implements BerkeleyAdditionalS
...
 
   private Database substateAccumulatorHashDatabase;
   private Database epochHashDatabase;
+  private Database lastEpochHashVerifiedDatabase;
 
   private byte[] currentSubstateAccumulatorHash;
   private Optional<Long> lastEpochInDbOpt;
   private Optional<Long> lastEpochInFileOpt;
   private Optional<Long> lastStateVersionInDbOpt;
+  private Optional<Long> lastEpochHashVerified;
```
siy, line 133: Perhaps it worth to consider use of `OptionalLong`.

siy (approved):
LGTM with one minor comment

radixdlt-core/radixdlt/src/main/java/com/radixdlt/statecomputer/substatehash/BerkeleySubstateAccumulatorHashStore.java
diegomrsantos: Oh thanks for the suggestion, I didn't know it. I will use it in the future, but does it offer any big advantage to justify changing it now?

siy: It eliminates boxing/unboxing and not susceptible to NPE. If highest possible performance is not a requirement, then there is no significant difference between `OptionalLong` and `Optional<Long>`.

PR #682, opened 2022-05-02
Substate change accumulator

diegomrsantos:


dhedey (approved):
Looks good, thanks Diego 👍

PR #690, opened 2022-05-18
Rework event queue cache in `EpochManager`

siy:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
```diff
@@ -342,26 +328,22 @@ private void processConsensusEventInternal(ConsensusEvent consensusEvent) {
...
-      queuedEvents
-          .computeIfAbsent(consensusEvent.getEpoch(), e -> new ArrayList<>())
-          .add(consensusEvent);
-      counters.increment(CounterType.EPOCH_MANAGER_QUEUED_CONSENSUS_EVENTS);
-      return;
-    }
+      if (consensusEvent.getEpoch() == this.currentEpoch() + 1) {
+
+        // Keep only events for the last view
+        if (isNotTheSameViewAs(consensusEvent)) {
```
dhedey, line 336: I'm not sure I understand what this is doing. It allow allows view events to be added for the same view as the first in the list. What if a view 2 event came in before the view 1 event?

Could we instead (eg) restrict ourselves to only tracking events for (eg) the first 20 rounds of the next epoch? (enough for a couple of timeouts). We could even store these in a (round -> event) map if we wished to

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
```diff
@@ -371,6 +353,11 @@ public void processConsensusEvent(ConsensusEvent consensusEvent) {
     this.processConsensusEventInternal(consensusEvent);
```
dhedey, line 353: What are your thoughts on maybe calling this `processConsensusEventForCurrentEpoch`?

dhedey (changes requested):
Not sure on this `isNotTheSameViewAs` as discussed in slack.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
siy: Actually, I've found that this version does not pass some integration tests. I've removed check and let the cache hold all events which belong to next epoch.

dhedey (approved):
Looks good.

When ready, could we also have this merged into main and babylon?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
```diff
@@ -312,63 +310,69 @@ private void processEpochChange(EpochChange epochChange) {
...
-        .forEach(this::processConsensusEventInternal);
-
-    queuedEvents.remove(epochChange.getEpoch());
+    this.processCachedConsensusEventsAtStartOfEpoch(epochChange, queuedEventsForNewEpoch);
   }
 
-  private void processConsensusEventInternal(ConsensusEvent consensusEvent) {
-    this.counters.increment(CounterType.BFT_EVENTS_RECEIVED);
+  public void processConsensusEvent(ConsensusEvent consensusEvent) {
+    final var currentEpoch = this.currentEpoch();
```
diegomrsantos, line 324: I believe we can just use `this.currentEpoch()` instead of creating a variable.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
```diff
@@ -312,63 +310,69 @@ private void processEpochChange(EpochChange epochChange) {
...
 
-    switch (consensusEvent) {
-      case Proposal proposal -> bftEventProcessor.processProposal(proposal);
-      case Vote vote -> bftEventProcessor.processVote(vote);
+    if (consensusEvent.getEpoch() == currentEpoch) {
```
diegomrsantos, line 326: I would use a `if`, `else if` and `else` in this function instead of 3 ifs.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
```diff
@@ -312,63 +310,69 @@ private void processEpochChange(EpochChange epochChange) {
...
           () -> consensusEvent,
-          this::currentEpoch);
-      return;
+          () -> currentEpoch);
     }
+  }
+
+  private void processCachedConsensusEventsAtStartOfEpoch(
+      EpochChange newEpochChange, List<ConsensusEvent> queuedEventsForNewEpoch) {
+    var newEpoch = newEpochChange.getEpoch();
```
diegomrsantos, line 348: Can we pass `newEpochChange.getEpoch()` as the parameter to the function?

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
dhedey: That's definitely a semantically correct choice. Although, to explain - this was a purposeful choice to follow the "early return" pattern - https://dev.to/jpswade/return-early-12o5

In general, I aim to reduce nesting and the amount of stuff you need to remember, and I typically avoid else blocks, favouring taking out methods and early returns; which I (personally) find to be a pattern which is more generally applicable and easier to read.

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
dhedey: This was a remnant from where I was concerned with race conditions - but I think this is still fine - as it means someone reading the method doesn't have to worry about `this.currentEpoch()` changing mid-method, which decreases the intellectual burden. (of course, if you know that all of this runs in a single thread, I guess this is irrelevant)

LukasGasior1 (approved):

radixdlt-core/radixdlt/src/main/java/com/radixdlt/consensus/epoch/EpochManager.java
diegomrsantos: Sonar says it is a major code smell. If it wasn't single-threaded, I'm not sure we should change this field arbitrarily.

diegomrsantos: Seems Sonar is complaining cause both variables have the same name.

radixdlt-core/radixdlt/src/integration/java/com/radixdlt/integration/steady_state/simulation/consensus_ledger_sync_epochs/RandomValidatorsTest.java
```diff
@@ -85,13 +85,13 @@
...
+  private final SyncConfig syncConfig = SyncConfig.of(200L, 10, 100L);
 
   private final Builder bftTestBuilder =
       SimulationTest.builder()
           .networkModules(NetworkOrdering.inOrder(), NetworkLatencies.fixed())
           .ledgerAndEpochsAndSync(
-              View.of(3),
+              View.of(10),
               goodRandomEpochToNodesMapper(),
               syncConfig) // TODO: investigate why this fails with View.of(10)
```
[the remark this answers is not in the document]
siy: Investigation is still necessary. My change resolved the issue, but we still don't know what caused it.

PR #695, opened 2022-05-20
Fix releasenet fork info

Theoklitos:
This is to make ReleaseNet compatible with the latest main branch.

The fork names are shortened due to
```
Caused by: java.lang.IllegalArgumentException: Fork name can't be longer than 16 bytes
```


dhedey (approved):
Looks good, thanks Theo!

PR #708, opened 2022-07-20
Upgrade base docker image to ARM-compatible, LTS Ubuntu version

KrzysztofLabus-RDX:
Upgrade base Ubuntu image to most recent ARM-compatible, LTS version.

PR #711, opened 2022-07-28
Ubuntu base ubuntu image upgrade & publish amd64/arm64 images

KrzysztofLabus-RDX:
This is improved version of previous PR https://github.com/radixdlt/radixdlt/pull/708

1. Use Ubuntu 22.04 LTS as base image for all radixdlt-core images.
2. Use Ubuntu 22.04 LTS as running environment for GitHub action.
3. Build and publish for linux-x86 (amd64) and linux-arm64 (arm64) architectures.

As of right now use of Docker version `20.10.10` or newer is required!


.github/workflows/build-publish-jars-workflow.yml
```diff
@@ -18,7 +18,7 @@ jobs:
   publish-jar:
     environment: publish-artifacts
     name: Publish JAR to Github package repository
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
shambupujar, line 21: I wasn't sure whether the jar that we package should be using 22.04 as node-runners would be using 20.04  as per docs [here](https://docs.radixdlt.com/main/node-and-gateway/node-setup-introduction.html#_prerequisites) 

However I see we previously been using latest. So it may be okay as no one reported issue so far

KrzysztofLabus-RDX: Should we resolve this thread then or do you want to do some extra tests for this action?

radixdlt-core/docker/scripts/fix-vulnerabilities.sh
```diff
@@ -22,6 +6,3 @@ apt remove -y libglib2.0-0 libglib2.0-bin libglib2.0-data
 
 #CVE-2013-7445 and others (neded to compile drivers)
 apt remove -y linux-libc-dev
-
-apt remove -y unzip wget
-apt autoremove -y
```
dhedey, line 27: We're no longer updating stuff (is that why we've moved to `RUN apt update -y` in the Docker file?); We're also no longer removing unzip, curl, wget etc. This potentially introduces attack surfaces.

Why was this change made? Is there something different about Ubuntu 22.04 which suggests we don't need this?

dhedey (commented):
Quick check before merging about increased attack surface inside the docker container

radixdlt-core/docker/scripts/fix-vulnerabilities.sh
KrzysztofLabus-RDX: 1. I've moved `apt get update -y` from this `fix-vulnerabilities.sh` to `Dockerfile` just to improve maintainability. Previously we were relying on side-effects of this script in our Dockerfile (java sdk installation relied on this apt update). Same is true for `autoremove`.
2. I've removed `remove unzip wget` as it was only needed for this manual `curl` installation which is no longer needed as we can rely on `curl` shipped with Ubuntu 22.04 - it's no longer vulnerable to this CVE.

.github/workflows/build-publish-jars-workflow.yml
shambupujar: Yeah should be fine. I will add a test to ticket where I need to do changes for CLI

shambupujar (approved):

dhedey (approved):
Looks good to me! Thanks!

PR #726, opened 2023-08-08
Use a separate http server for the "end state" API

LukasGasior1:


radixdlt-core/radixdlt/src/main/java/com/radixdlt/RadixNodeModule.java
```diff
@@ -137,8 +137,11 @@ public final class RadixNodeModule extends AbstractModule {
           Network.SANDPITNET.getId(), new GenericTestnetForksModule(),
           Network.LOCALNET.getId(), new GenericTestnetForksModule());
 
-  private static final int DEFAULT_CORE_PORT = 3333;
-  private static final String DEFAULT_BIND_ADDRESS = "0.0.0.0";
+  private static final int DEFAULT_PRIMARY_API_PORT = 3333;
+  private static final String DEFAULT_PRIMARY_API_BIND_ADDRESS = "127.0.0.1";
```
malexandru-rdx, line 141: Is the default change from `0.0.0.0` to `127.0.0.1` intended?

malexandru-rdx (approved):
LGTM!

radixdlt-core/radixdlt/src/main/java/com/radixdlt/RadixNodeModule.java
LukasGasior1: Yes, core API really shouldn't be exposed externally (unless one knows what they're doing)

LukasGasior1: Yes. And the other one is `RADIXDLT_END_STATE_API_BIND_ADDRESS`

